PR #22, opened 2016-01-08
Fix #21 - First cut of terracotta-kit

### LINKED ISSUE #21: Create a server package including all dependencies and a run script.
After internal discussions, we came up with following structure for server packaging, this issue is for creating a kit with below structure. 

<pre>
server-kit/
      ├── plugins
              ├── api
              └── lib
      ├── legal
      ├── server
              ├── bin
              └── lib
</pre>

ramsai1729:
#### Changes
* a terracotta-kit with structure mentioned in the issue #21 
* Changes to class loading to support the new server structure 

#### Few things to note
* need a license to start a server, we need to discuss further and fix this 
* new class loading changes triggered using system property "using-terracotta-kit"


terracotta-kit/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+
+    <artifactId>terracotta-kit</artifactId>
+    <packaging>pom</packaging>
+    <name>Terracotta Kit</name>
+
+
+    <dependencies>
+        <dependency>
+            <groupId>org.terracotta</groupId>
+            <artifactId>terracotta</artifactId>
```
ljacomet: What is this dependency? Don't think we have such an artifact in the 5 line

common/src/main/java/com/tc/config/Directories.java
```diff
@@ -81,4 +81,12 @@ public static File getServerLibFolder() throws FileNotFoundException {
     return new File(getInstallationRoot(), "server/lib");
   }
 
+  public static File getServerPluginsApiDir() throws FileNotFoundException {
+    return new File(getInstallationRoot(), "plugins/api");
```
myronkscott: Will this be an issue on windows?

myronkscott: File.separator?

common/src/main/java/com/tc/classloader/ServiceLocator.java
```diff
@@ -116,4 +115,44 @@
     return implementations;
   }
```
myronkscott: Let's try and explore some ways we can do this part differently.  Using system properties and directory scanning may be OK in the interim but I'd like to find a more sustainable solution.

myronkscott:
I like that you separated out the common component checking.  We should refine the implementation of the initial checks that we do before we merge.

common/src/main/java/com/tc/config/Directories.java
ramsai1729: not a problem on windows

terracotta-kit/pom.xml
ramsai1729: yeah, this dependency is causing the build failure. 

This particular module is part of integration-root and used by forge plugin to copy server dependencies into kit. we have few options 

1. try using mvn dependency plugin to copy dependencies 
2. create a terracotta module in OSS
3. modify forge plugin

I will try these option in order.

common/src/main/java/com/tc/classloader/ServiceLocator.java
ramsai1729: I am open for ideas.

terracotta/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>terracotta-core-root</artifactId>
+        <groupId>org.terracotta.internal</groupId>
+        <version>5.0.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>terracotta</artifactId>
```
ljacomet: Why do we need this new module?
Just for the dependency resolution?
How can this not be achieved by using the only dependency declared here as the kit main dependency?

ramsai1729: As mentioned in the email thread, this module is used for below reasons: 

* terracotta package has required server class path in manifest and used in start scripts.
* used for copying dependencies

Could you add mvn "install" phase to cloudbees build job to get the build passing? Let me know if we need to create a issue for this.

terracotta-kit/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+            <version>${project.version}</version>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.terracotta</groupId>
+                <artifactId>maven-forge-plugin</artifactId>
```
ljacomet: No idea why this plugin requires the artifact to be installed.
If replaced with

```
<groupId>org.apache.maven.plugins</groupId>
<artifactId>maven-dependency-plugin</artifactId>
<version>2.10</version>
```
and after adapting the configuration - that is removing everything but the output dir which becomes `outputDirectory`, it builds fine with just the verify task.

terracotta/pom.xml
ljacomet: See other comment on this topic

ljacomet:
Had to kill the previous build - was stuck - please have a look.

Restarting one though

Test this please

ramsai1729:
looks like build was stuck due to some other reason, not this PR. Anyways its passing now.

myronkscott:
We have been having a terrible time with stuck builds and JVM kills on jenkins with no obvious cause.  Without any evidence for test failures or test hangs, I've been assuming these are infrastructure related and excluding machines that fail to build tc-core.

PR #53, opened 2016-02-01
Config parsing with classloaders

myronkscott:
The configuration parser needs a special class loader so that the services that are loaded by the kit use the correct class hierarchy of the platform.  This implementation feeds the class loaders used by the Service Locator to  configuration.


myronkscott:
Tests needed.

PR #72, opened 2016-02-08
Fix #66 - remove ServiceLoader support for platform services loading

### LINKED ISSUE #66: Remove ServiceLoader support for Platform services loading
Looks like this might break after recent class loading changes. Its better to remove this as we are not going to support anyway.

ramsai1729:


dso-l2/src/main/java/com/tc/services/TerracottaServiceProviderRegistryImpl.java
```diff
@@ -53,27 +53,12 @@ public void initialize(String serverName, TcConfiguration configuration) {
             registerNewServiceProvider(provider);
           }
         } catch (InstantiationException | IllegalAccessException ie) {
-//  really shouldn't be doing this.  ServiceProvider configurations should provide appropriate classes
-//  to handle the config.
-          tryServiceLoader(config);
+          logger.error("caught exception while initializing service " + serviceClazz, ie);
```
myronkscott: Should this re-throw and crash the server or just fail to load the service?  I'm inclined to say this should re-throw and crash.  Fail fast.

jd0-sag: This is a good idea:  I am worried that this is a serious problem which we are going to ignore and relegate to logs.

We should re-throw, probably.  I would want to see the server fail start-up.

PR #75, opened 2016-02-09
MGMT key deferred execution

myronkscott:
This PR makes MGMT_KEY more exclusive than before.  MGMT_KEY currently flushes all the request queues but does not prevent subsequently submitted requests from overrunning the event on the MGMT_KEY.  The provides a deferment queue so that new messages on this queue are not submitted request processor until after the MGMT_KEY event completes.

Also included in this PR is a bug fix in the passive execution pipeline.


jd0-sag:
These generally look good.  There are a few extraneous changes which seem to waffle back and forth, which should probably be removed, and some of these new ideas could warrant some greater documentation but it otherwise looks good to go.

PR #80, opened 2016-02-10
Fix #78: Updated EntityClientCommunicatorService to use EntityResponse as its argument type

### LINKED ISSUE #78: Update EntityClientCommunicatorService to reflect API change
This is the corresponding core implementation to the change in API described in Terracotta-OSS/terracotta-apis#12.

jd0-sag:
-also updated the corresponding test to use this type and provide a mock codec for serialization


jd0-sag:
Note that this change depends on Terracotta-OSS/terracotta-apis#46 and they should be merged together.

PR #81, opened 2016-02-11
Log exceptions during request processing

ramsai1729:


dso-l2/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
```diff
@@ -335,6 +336,7 @@ private void invoke(ServerEntityRequest request, EntityMessage message, int conc
       } catch (Exception e) {
         // Wrap this exception.
         EntityUserException wrapper = new EntityUserException(id.getClassName(), id.getEntityName(), e);
+        logger.error("caught exception during invoke ", wrapper);
         request.failure(wrapper);
         throw new RuntimeException(e);
```
ramsai1729: should we throw this exception?

jd0-sag:
We need to discuss this in tonight's meeting.

dso-l2/src/main/java/com/tc/objectserver/handler/ProcessTransactionHandler.java
```diff
@@ -193,7 +196,12 @@ private synchronized void addMessage(ClientID sourceNodeID, EntityDescriptor des
       }
     } else {
       // If there was an exception of any sort, just pass it back as a failure.
+      logger.error("caught exception while handling the request ", uncaughtException);
       serverEntityRequest.failure(uncaughtException);
+      // crash the server if exception is a EntityUserException
+      if(uncaughtException instanceof EntityUserException) {
```
jd0-sag: Why are we checking the exception type?  Can we throw when we first create/catch it, instead of at this later point?

ramsai1729: this particular `uncaughtException` is from 165 line and we need to persist the transaction before re-throwing the exception.

jd0-sag: My thinking is that this should actually just be a different variable since we want to treat it very differently.

In general (and there are a few exceptions to this rule), an explicit instanceof type check means that something else is wrong, in the design.

ramsai1729:
Rebased.

dso-l2/src/main/java/com/tc/objectserver/handler/ProcessTransactionHandler.java
ramsai1729: Introduced a new local variable.

PR #92, opened 2016-02-22
Fix #91: crash the server if entity deserializer returns null

### LINKED ISSUE #91: Request hangs if implementation of MessageCodec.deserialize(byte[] payload) returns null
Looks like we are not sending any reply to the client if  MessageCodec.deserialize(payload) return null. After discussions we have decided to crash the server if MessageCodec.deserialize(payload) as its not supposed to return null.

ramsai1729:
This PR brings changes to crash the server if entity MessageCodec.deserialize(...)/deserializeForSync(...) returns null.


ramsai1729:
Rebased.

PR #94, opened 2016-02-23
Remove slf4j implementation from the shaded jar

anthonydahanne:
It's up to the consumers of the client runtime library to provide their slf4j implementation.
In a project using client-runtime, I'm currently encountering slf4j issues since it gets 2 impl. in the classpath.

PR #100, opened 2016-02-25
Reworked TCLogger and TCLogging / Replaces PR #94

mathieucarbou:
This PR replaces PR #94. It includes @anthonydahanne commits, and changes on the TCLogger and TCLogging to support SLF4J and hides Log4J for client runtime. 

Spoken to @lorban and @anthonydahanne : this is a first commit which goal is to make the client-runtime lib __usable__ in client projects. There will still need to completely rework this logging system.

    Updated TCLogging to remove the hard log4j dependencies
    
     - Supports Log4j  by default if it is in the classpath, otherwise uses SLF4J
     - API break: had to remove TCLogger.setLevel() method and some other public methods on TCLogger
     - Provided implementations of TCLogger and TCLogging for both Log4j and SLF4J
     - Removed the JMX Appender (confirmed with @lorban)
     - Client Runtime only contains SLF4J api and ships with externalized log4j and slf4j-log4j
    
- All tests pass
- start-tc-server.sh starts correctly
- I tried Client Runtime in our app, our app can start and client runtime logs can be seen

__IMPORTANT_

This is possible that this PR causes compile issues on other projects relying on TCLogging or TCLogger methods that have been removed. In such case, client code should be updated because we cannot rely on log4j-dependant classes, at least for client-runtime.


mathieucarbou:
@anthonydahanne, @lorban : this is the PR we talked about.
@ramsai1729 : it replaces PR #94.

jd0-sag:
I am having trouble understanding the scope of the last commit since it includes too many conflated ideas.  Can that be split up into a sequence of commits, so the ideas are isolated?

Also, there is this bizarre removal of the word "public" from many classes, which is extraneous (and sometimes changes the license header).  This part of the change should be reverted.

mathieucarbou:
@jd0-sag : I updated the PR so that is is easier for your team to review.

I firstly renamed classes and made them package local to isolate classes that depends on specific log4j stuff and are only used by the `Log4jTCLogging` class.

I've made this change to be sure that no module will ever use these classes that could cause a client crash in the case a client is not using log4j.

ramsai1729:
I have added few cleanup comments, otherwise this PR seems fine to me. Also, lets not change visibility of the classes.

mathieucarbou:
@ramsai1729 : thanks! I'll update. But please see with @jd0-sag and @myronkscott before merging :-)

mathieucarbou:
Rebased!

myronkscott:
Obsoleted by PR#106

PR #105, opened 2016-03-01
Fix #104 - halt processing after a MessageCodecException

### LINKED ISSUE #104: Do not process an EntityMessage after a MessageCodecException
If a MessageCodecException is returned to the client, no further processing can take place.  Currently, the server is incorrectly crashed.

myronkscott:
The exception is returned to the client.  Halt processing on the server after exception.

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
@@ -146,6 +146,8 @@ public class ManagedEntityImpl implements ManagedEntity {
         message = runWithHelper(()->codec.deserialize(payload));
       } catch (EntityUserException e) {
         request.failure(e);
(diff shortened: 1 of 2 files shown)


ramsai1729:
Okay, but don't we need to **crash** the server **without** sending the response to client even if it is **MessageCodecException**? I thought that was our decision.

myronkscott:
I'm a bit confused as well.  I thought we decided that returning null from a MessageCodec was not supported, crash the server.  But if a MessageCodecException was thrown, then that was sent back to the client and processing was stopped.  

I think we decided that it was ok to throw an EntityUserException inside entity code in invokes and MessageCodec.  These would be exceptions that entity implementors could catch and account for in either Entity client or server code.  It's the uncaught exceptions inside entity code that should bring down the server.

We should probably discuss again.

myronkscott:
This is an improper fix.  Closing to re-think fix.

PR #125, opened 2016-03-15
Fix #5 - Follow up changes for Service API change

### LINKED ISSUE #5: Services must adhere to dirtydb
State machine in core is a hold-over from the previous version of the platform.  In that version, storage was directly integrated in the system and the platform could directly clear the persistent data as a result of a dirtydb mark and zap signal.  Going forward, storage will be implemented as a service and storage is no longer directly integrated into platform.  As a result, storage services will need a mechanism to know about dirtydb so that any persistent state can be cleared.

ramsai1729:
Implement clear() method for all ServiceProvider and IPersistentStorage implementations
Clear all Services state if dirty db is set

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/persistence/FlatFileStorageServiceProvider.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/persistence/FlatFileStorageServiceProvider.java
@@ -21,6 +21,7 @@ package com.tc.objectserver.persistence;
 
 import org.terracotta.entity.ServiceProvider;
 import org.terracotta.entity.ServiceConfiguration;
+import org.terracotta.entity.ServiceProviderCleanupException;
 import org.terracotta.persistence.IPersistentStorage;
 
 import com.tc.logging.TCLogger;
@@ -32,6 +33,9 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Set;
+import java.util.HashSet;
+
 import org.terracotta.entity.ServiceProviderConfiguration;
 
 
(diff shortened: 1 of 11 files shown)


dso-l2/src/main/java/com/tc/objectserver/persistence/FlatFileStorageServiceProvider.java
```diff
@@ -72,7 +77,9 @@ public boolean initialize(ServiceProviderConfiguration configuration) {
     if (!this.shouldPersistAcrossRestarts) {
       file.delete();
     }
-    return configuration.getServiceType().cast(new FlatFilePersistentStorage(file));
+    FlatFilePersistentStorage storage = new FlatFilePersistentStorage(file);
+    storages.add(storage);
```
jd0-sag: Will this be sufficient?  If we try to clear() before all the entities have started (which is always the case), their files will still be on disk.

Or do we ensure that these are deleted when the entity is first created (or otherwise first tries to open the file)?

dso-l2/src/main/java/com/tc/objectserver/persistence/FlatFileStorageServiceProvider.java
```diff
@@ -84,5 +91,18 @@ public boolean initialize(ServiceProviderConfiguration configuration) {
   public void close() {
     
   }
+
+  @Override
+  public void clear() throws ServiceProviderCleanupException {
+    for(FlatFilePersistentStorage storage : storages) {
```
jd0-sag: I have a feeling that this will actually have to walk the underlying directory, deleting any files it finds, since we have not opened them all, yet.

dso-l2/src/main/java/com/tc/objectserver/persistence/Persistor.java
```diff
@@ -64,8 +65,8 @@ public Persistor(IPersistentStorage persistentStorage) {
   public void start() {
     sequenceManager = new SequenceManager(persistentStorage);
     clientStatePersistor = new ClientStatePersistor(sequenceManager, persistentStorage);
-
-    if (!this.clusterStatePersistor.isDBClean()) {
+    wasDBClean = this.clusterStatePersistor.isDBClean();
```
jd0-sag: Do we still want to do this, here, or can we rely on the server forcing us into a clean state, from the top?

PR #127, opened 2016-03-18
Move plugins folder into the server folder (kit)

anthonydahanne:
The plugins folder is inherent to the server, since thanks to the jars added to the plugins directory, the server can "understand" new entities and services.
Therefore, the plugins are inherent to the server.

It is confusing to see the plugins folder at the same level as the server folder : one could there are separated concepts; hence the move.

Moreover, when included in other kits, it is hard to understand what "plugins" stand for, if no in the context of the server.

### DIFF
--- a/terracotta-kit/src/assemble/server/bin/start-tc-server.sh
+++ b/terracotta-kit/src/assemble/server/bin/start-tc-server.sh
@@ -45,8 +45,8 @@ trap cleanup_INT SIGINT
 
 THIS_DIR=`dirname $0`
 TC_INSTALL_DIR=`cd $THIS_DIR;pwd`/../..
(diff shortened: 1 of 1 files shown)


myronkscott:
What about the client side of entities?  Some of our test entire package everything in one jar.  Even if you did separate jars, where would you put the client side of entities?

anthonydahanne:
Well, I believe the client jars are there for the user to add to his classpath.
Using ehcache3 clustered ? get the uber jar with everything included
Using your own entities ? get the jar (client-runtime) you provide in the client folder of the kit.

myronkscott:
We are going to defer this until plugins folder is better settled.  Currently some entities package as single jar making move to server folder improper.

PR #135, opened 2016-03-25
Fix Destroy corner cases

myronkscott:
Destroy during sync and handshake pose problems because they can leave a client or passive in an inconsistent state if the entity disappears during an operation.  Fix these cases by special handling and and checking for the destroy case.


myronkscott:
Not properly validated.

PR #142, opened 2016-04-01
reset sent requests semaphore on reconnects

myronkscott:
semaphore permits can get drained for sent requests that get lost during reconnect


myronkscott:
self merging to fix build pipeline

PR #143, opened 2016-04-01
Fix #137 - cleanup tc properties

### LINKED ISSUE #137: tc.properties - Replace string literals usage with named constants
Few examples: 

dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java-                                                               new HealthCheckerConfigImpl(tcProperties
dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java:                                                                   .getPropertiesFor("l2.healthcheck.l1"), "TSA Server"),
dso-l2/src/main/java/com/tc/net/utils/L2Utils.java:    return TCPropertiesImpl.getProperties().getInt("l2.tccom.workerthreads", def);
dso-l2/src/main/java/com/tc/net/utils/L2Utils.java:    return TCPropertiesImpl.getProperties().getInt("l2.seda.stage.workerthreads", def);
dso-l2/src/test/java/com/tc/config/TcPropertiesWithSpacesOverWriteTest.java:    Assert.assertEquals(false, tcProps.getBoolean("l2.l1reconnect.enabled"));
dso-l2/src/test/java/com/tc/config/TcPropertiesWithSpacesOverWriteTest.java:    Assert.assertEquals(10000, tcProps.getInt("l1.socket.connect.timeout"));
dso-l2/src/test/java/com/tc/config/TcPropertiesWithSpacesOverWriteTest.java:    Assert.assertEquals(2000, tcProps.getInt("l1.socket.reconnect.waitInterval"));
dso-l2/src/test/java/com/tc/net/groups/TCGroupManagerImplTest.java:    long minTimeToSayDead = TCPropertiesImpl.getProperties().getLong("l2.healthcheck.l1.ping.idletime")
dso-l2/src/test/java/com/tc/net/groups/TCGroupManagerImplTest.java:                            + (TCPropertiesImpl.getProperties().getLong("l2.healthcheck.l1.ping.interval") * TCPropertiesImpl
dso-l2/src/test/java/com/tc/net/groups/TCGroupManagerImplTest.java:    minTimeToSayDead += (3 * TCPropertiesImpl.getProperties().getLong("l2.healthcheck.l1.ping.interval"));
dso-l2/src/test/java/com/tc/net/groups/TCGroupManagerImplTest.java:      ThreadUtil.reallySleep(TCPropertiesImpl.getProperties().getLong("l2.healthcheck.l1.ping.interval")
dso-l2/src/test/java/com/tc/net/groups/TCGroupManagerImplTest.java:                             * TCPropertiesImpl.getProperties().getLong("l2.healthcheck.l1.ping.probes"));

ramsai1729:
* replace tc properites string literals usage with named constants
   * remove unnecessary public static final usage in TCPropertiesConsts interface
   * create subcategories for use with TCProperties.getPropertiesFor(...)

### DIFF
--- a/common/src/main/java/com/tc/net/core/SocketParams.java
+++ b/common/src/main/java/com/tc/net/core/SocketParams.java
@@ -22,6 +22,7 @@ import com.tc.logging.TCLogger;
 import com.tc.logging.TCLogging;
 import com.tc.properties.TCProperties;
 import com.tc.properties.TCPropertiesImpl;
+import com.tc.properties.TCPropertiesConsts;
 
 import java.net.ServerSocket;
 import java.net.Socket;
@@ -33,7 +34,6 @@ import java.net.SocketException;
 class SocketParams {
   private static final TCLogger logger       = TCLogging.getLogger(SocketParams.class);
 
-  private static final String   PREFIX       = "net.core";
   private static final String   RECV_BUFFER  = "recv.buffer";
   private static final String   SEND_BUFFER  = "send.buffer";
   private static final String   TCP_NO_DELAY = "tcpnodelay";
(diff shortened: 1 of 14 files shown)


jd0-sag:
Don't remove those public static final decorations.  They may be redundant in the case of an interface but people shouldn't have to realize that they are looking at an interface (in fact, TCPropertiesConsts being an interface is already peculiar).  Marking them as such makes it clear what is going on (otherwise, they _look_ like mutable instance fields with initialized values).  It also creates needless churn in the version history.

Also, the imports should probably be of the class, not the individual constants.  That means that you only import once and also makes it clear where the constant comes from, just by looking at it (without the class name, it first appears to be a local declaration).

ramsai1729:
Rebased, please check and merge.

PR #156, opened 2016-04-14
Election fix

myronkscott:
This PR fixes an issue with elections and restartable servers.  Currently, restartable servers can get caught in an endless election loop with passive-standbys.  This fixes that issue by participating in an election whenever a server receives a signal from a peer that an election is starting.  An additional stage was created to support this new election semantic.  Numerous test fixes included.


myronkscott:
test this please

myronkscott:
test this please

Issue #158, opened 2016-04-18
Create start up script for windows (start-tc-server.bat)


anthonydahanne:
Windows users can't use the start-tc-server.sh script.
They would need a start-tc-server.bat

PR #172, opened 2016-05-03
initial import

gibsong:
This pull request is for issue #158 
https://github.com/Terracotta-OSS/terracotta-core/issues/158


jd0-sag:
I should have merged this before another change that I merged, this morning, so can you rebase this and then I will merge it?

Also, if you add the text "Fix #158" to the commit comment (we usually put it in the first line, so we quickly see context) then the commit will auto-close the issue when it merges.  This isn't essential but, if you feel like rewording the commit during the rebase, it will handle the close for you.

Either way, I can merge after the rebase.

PR #176, opened 2016-05-05
Active-passive parallel invoke support

jd0-sag:
* this uses a new RECEIVED ack added in the active-passive dialect (note that this is NOT exposed to the client)
* the OLD logic on the active was:  block until passive completes, now run invoke, then return
* the new logic on the active is:  block until passive receives, now run invoke, then block until passive completes, then return
* there are some concerns that exclusive mode might have been relying on some nature of the old ACK ordering but we believe that has changed and none of the tests had a problem
* by injecting a 20ms sleep between the "block for receives" and "invoke", I was able to observe MessageFloodTest running in both orders over the course of a run

### DIFF
--- a/tc-messaging/src/main/java/com/tc/l2/msg/ReplicationMessage.java
+++ b/tc-messaging/src/main/java/com/tc/l2/msg/ReplicationMessage.java
@@ -61,6 +61,21 @@ public class ReplicationMessage extends AbstractGroupMessage implements OrderedE
     SYNC_ENTITY_CONCURRENCY_END;
   }
   
+  // Factory methods.
+  public static ReplicationMessage createStartMessage() {
+    return new ReplicationMessage(START);
+  }
+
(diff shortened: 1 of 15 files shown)


tc-messaging/src/main/java/com/tc/l2/msg/ReplicationMessage.java
```diff
@@ -37,10 +37,11 @@
  */
 public class ReplicationMessage extends AbstractGroupMessage implements OrderedEventContext {
 //  message types  
-  public static final int REPLICATE               = 0; // Sent to replicate a request on the passive
-  public static final int SYNC               = 1; // Sent to replicate a request on the passive
-  public static final int RESPONSE                = 2; // response that the replicated action completed
-  public static final int START                = 3; // response that the replicated action completed
+  public static final int INVALID               = 0; // Sent to replicate a request on the passive
```
myronkscott: comment is wrong

PR #179, opened 2016-05-11
Fix #178: Fixed usage of paths with spaces

### LINKED ISSUE #178: `start-tc-server.sh` fails to handle spaces in install directory path
The present CloudBees-based build system for Ehcache performs the builds in a JDK-specific work directory.  Each of the working directories has at least one space in at least one directory name -- for example, `/scratch/jenkins/workspace/ehcache3-pull-requests/jdk/Sun JDK 1.6 (latest)/label_exp/standard`.  The builds are failing (hanging) in clustered:integration-test for tests dependent on starting a server (using `start-tc-server.sh`) -- the `PLUGIN_CLASSPATH` is not being properly built when the install directory contains a space.

This failure is presently blocking successful builds of PR ehcache/ehcache3#1039.

jd0-sag:
I will get someone to verify this, on Windows, before we can merge.

### DIFF
--- a/terracotta-kit/src/assemble/server/bin/start-tc-server.sh
+++ b/terracotta-kit/src/assemble/server/bin/start-tc-server.sh
@@ -69,17 +69,17 @@ function setPluginClasspath {
 
     shopt -s nullglob
 
-    for pluginDir in ${PLUGIN_LIB_DIR} ${PLUGIN_API_DIR}
+    for pluginDir in "${PLUGIN_LIB_DIR}" "${PLUGIN_API_DIR}"
     do
-        if [ -e ${pluginDir} ]
+        if [ -e "${pluginDir}" ]
         then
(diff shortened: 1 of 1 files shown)


jd0-sag:
@gibsong was able to verify that the change isn't required on Windows, due to a difference in how shell variables are treated.

This is ready to merge if you are good with it, @cljohnso

cljohnso:
Looks good -- there's a side issue with an "empty" element in the generated class path but that's an item for another issue.

I was planning on merging this but stopped -- the status says "Some checks haven't completed yet" but there's no link to a build to observe ... is there something else preventing the "check" from completing?

jd0-sag:
I assume that a build somewhere is just stuck.  The checks probably wouldn't do anything to verify the kit as all the tests for it are in other projects (which is something we still need to figure out how to address).

If you are happy with the change, go ahead and merge.

PR #183, opened 2016-05-13
Fix windows file names.  TAB-6880

myronkscott:
character replacement for windows platforms on lock directory generation


dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
```diff
@@ -406,8 +406,16 @@ public void callbackOnExit(CallbackOnExitState state) {
...
     final boolean restartable = l2DSOConfig.getRestartable();
-
-    final TCFile location = new TCFileImpl(new TCFileImpl(this.configSetupManager.commonl2Config().dataPath()), this.configSetupManager.dsoL2Config().serverName());
+    
+    TCFile dataLoc = new TCFileImpl(this.configSetupManager.commonl2Config().dataPath());
+    String serverName = this.configSetupManager.dsoL2Config().serverName();
+//  this is character replacement for windows platform file names.  This is probably a bogus way to 
+//  handle this.  Re-evaluate the way data directories are managed for 5.0 and fix this when a plan is
+//  devised
+    serverName = serverName.replace('.', '-');
```
ramsai1729: how about replacing with hex string representation?

PR #185, opened 2016-05-17
Fix #184 - Implement PlatformService

### LINKED ISSUE #184: Phase 1 - Create a Service to dump Platform state into logs
- Create a Platform built-in service to dump Server state in logs (like existing system test does) so that an entity (like M&M entity) can use this service to dump server state
- Create an entity that can used by galvan framework to dump the server state

See #154 for more details

ramsai1729:
* Create PlatformServiceImpl, PlatformServiceProvider and PlatformServiceConfiguration
  * Dump Platform State using dumphandler and serverbuilder

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
@@ -21,6 +21,7 @@ package com.tc.objectserver.impl;
 import com.tc.async.api.AbstractEventHandler;
 import com.tc.async.api.EventHandlerException;
 
+import com.tc.services.PlatformServiceProvider;
 import org.terracotta.entity.ServiceConfiguration;
 import org.terracotta.entity.ServiceRegistry;
 import org.terracotta.monitoring.IMonitoringProducer;
@@ -572,6 +573,9 @@ public class DistributedObjectServer implements TCDumper, LockInfoDumpHandler, S
     CommunicatorService communicatorService = new CommunicatorService(channelManager);
     serviceRegistry.registerBuiltin(communicatorService);
     final Stage<ServerEntityResponseMessage> communicatorResponseStage = stageManager.createStage(ServerConfigurationContext.SERVER_ENTITY_MESSAGE_RESPONSE_STAGE, ServerEntityResponseMessage.class,  new CommunicatorResponseHandler(communicatorService), 1, maxStageSize);
+
+    PlatformServiceProvider platformServiceProvider = new PlatformServiceProvider(this);
+    serviceRegistry.registerBuiltin(platformServiceProvider);
 
     // Creating a stage here so that the sink can be passed
     final Stage<LockResponseContext> respondToLockStage = stageManager.createStage(ServerConfigurationContext.RESPOND_TO_LOCK_REQUEST_STAGE, LockResponseContext.class, new RespondToRequestLockHandler(), 1, maxStageSize);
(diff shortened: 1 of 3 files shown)


dso-l2/src/main/java/com/tc/services/PlatformServiceConfiguration.java
```diff
@@ -0,0 +1,14 @@
+package com.tc.services;
+
+import org.terracotta.entity.ServiceConfiguration;
+import org.terracotta.monitoring.PlatformService;
+
+/**
+ * @author vmad
+ */
+public class PlatformServiceConfiguration implements ServiceConfiguration<PlatformService> {
```
jd0-sag: Are you able to use BasicServiceConfiguration for this (we should probably do a pass over our implementations since most just return a constant).

dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
```diff
@@ -241,6 +242,7 @@
 import com.tc.objectserver.handler.ServerManagementHandler;
 import com.tc.operatorevent.TerracottaOperatorEvent;
 import com.tc.operatorevent.TerracottaOperatorEventCallback;
+
```
jd0-sag: Probably want to make sure that there aren't any gratuitous formatting changes, in the import list.

dso-l2/src/main/java/com/tc/services/PlatformServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import org.terracotta.monitoring.PlatformService;
+
+/**
+ * @author vmad
+ */
+public class PlatformServiceImpl implements PlatformService {
+
+    private final DistributedObjectServer.DumpHelper dumpHelper;
+
+    public PlatformServiceImpl(DistributedObjectServer.DumpHelper dumpHelper) {
```
jd0-sag: If this type is being referenced from within the DistributedObjectServer, you probably want to float it out into its own file.

This kind of approach typically makes sense if the inner class isn't referenced from outside (it is just the implementation of an interface, for example) or if it is something directly associated with the class it is inside (an interface required to interact with it, an enum defining its constants, etc).

In this case, it is more like a cyclic dependency so it should probably be split out.

dso-l2/src/main/java/com/tc/services/PlatformServiceProvider.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Collections;
+
+/**
+ * @author vmad
+ */
+public class PlatformServiceProvider implements BuiltInServiceProvider {
+
+    private final PlatformServiceImpl platformService;
+
+    public PlatformServiceProvider(DistributedObjectServer.DumpHelper dumpHelper) {
```
jd0-sag: This wouldn't need to know about so many things if you just passed in a PlatformService object, creating it from outside.

jd0-sag:
Mostly small concerns.  Convenient that you managed to use the existing TCDumper, where possible.

myronkscott:
Looks good.  We need to tackle dumping clients and passives immediately so please create issues for generating those.

ramsai1729:
dumps on passives can be generated by processing invoke request, please check test entity in integration 

I will create a issue for handling client dumps, seems bit complicated.

dso-l2/src/main/java/com/tc/services/PlatformServiceConfiguration.java
ramsai1729: Yeah, BasicServiceConfiguration can be used. will remove this class.

PR #189, opened 2016-05-20
Fix to ActivePassiveAckWaiters handling

jd0-sag:
* Fix a leak where we never removed these waiters
* Don't assert that the waiter was waiting for the completed node in the case where the node unexpectedly disappeared


jd0-sag:
@esebasti Feel free to merge this if it resolves the issue you found.  Let me know if you need me to set up a test environment to verify your case or if you are able to use the one you already have.

PR #195, opened 2016-05-30
Issue 187

ramsai1729:


dso-l2/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
```diff
@@ -231,6 +233,21 @@ private void processSyncMessage(ServerEntityRequest sync, byte[] payload, int co
     }
   }
 
+  @Override
+  public void dumpStateTo(StateDumper stateDumper) {
+    if(activeServerEntity != null) {
+      if(activeServerEntity instanceof StateDumpable) {
```
jd0-sag: Probably want to add some doc here to explain that this instanceof is not a hack but actually by design, since we are currently allowing that to be an optional "opt-in" kind of capability (if we make that mandatory, we will remove this check).

dso-l2/src/main/java/com/tc/services/LogBasedStateDumper.java
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * A Log based implementation of {@link StateDumper}, key-value mappings will be logged using {@link TCLogger}
+ *
+ * @author vmad
+ */
+public class LogBasedStateDumper implements StateDumper {
+
+  private static final TCLogger LOGGER = TCLogging.getDumpLogger();
+  private final ConcurrentMap<String, LogBasedStateDumper> instances = new ConcurrentHashMap<>();
+  private final ConcurrentMap<String, String> dumpState = new ConcurrentHashMap<>();
```
jd0-sag: In the future, we might want to make this just log, directly, instead of storing the state.

I expect this object to evolve, however, so I don't think we need to worry about its final shape, at the moment.

dso-l2/src/main/java/com/tc/services/TerracottaServiceProviderRegistryImpl.java
```diff
@@ -110,4 +112,21 @@ private void registerNewServiceProvider(ServiceProvider service) {
     logger.info("Initializing " + service);
     serviceProviders.add(service);
   }
+
+  @Override
+  public void dumpStateTo(StateDumper stateDumper) {
+    for (ServiceProvider serviceProvider : serviceProviders) {
+      if(serviceProvider instanceof StateDumpable) {
```
jd0-sag: Again, we probably want some comment here to explain that this is opt-in, for the moment.

dso-l2/src/main/java/com/tc/services/LogBasedStateDumper.java
ramsai1729: storing the state helps logging state in proper hierarchical way, this implementation logs state in one components state first and its children state.  Anyways this can be changed at any point of time without affecting API users.

PR #207, opened 2016-06-07
Fix #206 : Fix election issues

### LINKED ISSUE #206: Fix election issues when a node left cluster

ramsai1729:


dso-l2/src/main/java/com/tc/l2/state/StateManagerImpl.java
```diff
@@ -410,7 +410,7 @@ private synchronized void handleStartElectionRequest(L2StateMessage msg) throws
       groupManager.sendTo(msg.messageFrom(), abortMsg);      
     } else if (!electionMgr.handleStartElectionRequest(msg)) {
 //  another server started an election.  Unclear which server is now active, clear the active and run our own election
-      startElectionIfNecessary(syncdTo);
+      startElectionIfNecessary(ServerID.NULL_ID);
```
myronkscott: I think this is wrong.  we need to start election in this case and it will not happen with this parameter.

myronkscott:
Can we add a test for the multiple passive case somewhere?  A test with multiple passives and the active goes down during passive sync of one of the passives is also a hole in our testing.

myronkscott:
Will add tests.  Merging for now.

PR #217, opened 2016-06-21
Using bash instead of sh

akomakom:
start-tc-server.sh uses bash syntax (eg function) but uses sh as the interpreter.   /bin/sh != /bin/bash on many unix flavors (eg Ubuntu, where /bin/sh = dash).   

In other words, this script cannot work without bash, hence it should say so.


terracotta-org:

jd0-sag:
Good find.  Will merge this and we can see if it causes any issues in any of our testing environments but any failures there should have been found by the implicit syntax requirement.

PR #221, opened 2016-06-22
Fix #213 - terracotta-kit: Move plugins dir to server dir

### LINKED ISSUE #213: Update terracotta-kit structure
The current kit structure, in terms of folders is:
```
<root>
    /client
    /legal
    /plugins
        /api
        /lib
    /server
        /bin
        /lib
```

Proposal is to evolve it into:
```
<root>
    /client
    /legal
    /server
        /bin
        /lib
        /plugins
            /api
            /lib
```
that is moving the `plugins` structure inside `server` folder.

The initial location of `plugins` comes from the idea that we may package entities as single jar and thus require them on the server and client classpath at the same time.
Current non trivial entities seem to favour a packaging in multiple jars which makes this scheme irrelevant.
In addition, I believe that explaining to users that to construct their application classpath they need to look into `client` but also `plugins` is not ideal.

ramsai1729:
This PR creates terracotta-kit structure mentioned in #213 , basically moving plugins directory under server directory

new structure 
```
14:26:04 ~/Projects/work/kits $ tree -d terracotta-5.0.0-SNAPSHOT/
terracotta-5.0.0-SNAPSHOT/
├── client
├── legal
└── server
    ├── bin
    ├── lib
    └── plugins
        ├── api
        └── lib

8 directories
14:26:08 ~/Projects/work/kits $ 
```

### DIFF
--- a/terracotta-kit/src/assemble/distribution.xml
+++ b/terracotta-kit/src/assemble/distribution.xml
@@ -56,7 +56,7 @@ The Initial Developer of the Covered Software is
     </fileSet>
     <fileSet>
       <directory>${basedir}/src/assemble/plugins</directory>
-      <outputDirectory>/plugins</outputDirectory>
+      <outputDirectory>/server/plugins</outputDirectory>
     </fileSet>
     <fileSet>
       <directory>${basedir}/src/assemble/legal</directory>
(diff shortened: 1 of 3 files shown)


jd0-sag:
We should probably merge this last, out of our current sequence.  We also need to take a quick look to see if this has implications for galvan server installation.

myronkscott:
The plan was to rename "plugins" to "ext" in the move.  Seems more consistent with the overall directory scheme.

myronkscott:
Actually, I am wrong.  We are sticking with plugins

myronkscott:
The priority of this has been increased.  I rebased your commit for immediate merge and made the required galvan changes.

myronkscott:
this was done in a different PR y cherry-picking the commit on the current head

Issue #222, opened 2016-06-22
NPE during server startup on solaris


lorban:
Here's the stack trace:
```
java.lang.NullPointerException
        at com.tc.runtime.TCMemoryManagerImpl$MemoryMonitor.<init>(TCMemoryManagerImpl.java:126)
        at com.tc.runtime.TCMemoryManagerImpl.startMonitorIfNecessary(TCMemoryManagerImpl.java:105)
        at com.tc.runtime.TCMemoryManagerImpl.registerForMemoryEvents(TCMemoryManagerImpl.java:77)
        at com.tc.objectserver.impl.DistributedObjectServer.start(DistributedObjectServer.java:536)
        at com.tc.server.TCServerImpl.startDSOServer(TCServerImpl.java:446)
        at com.tc.server.TCServerImpl.access$300(TCServerImpl.java:77)
        at com.tc.server.TCServerImpl$StartAction.execute(TCServerImpl.java:423)
        at com.tc.lang.StartupHelper.startUp(StartupHelper.java:52)
        at com.tc.server.TCServerImpl.startServer(TCServerImpl.java:440)
        at com.tc.server.TCServerImpl.start(TCServerImpl.java:236)
        at com.tc.server.TCServerMain.main(TCServerMain.java:60)
```

myronkscott:
Fixed with PR #224

PR #224, opened 2016-06-22
#222 remove solaris-specific code working on null variable

lorban:


jd0-sag:
Do we have any idea why this special-case was here?  Removing it seems like the right answer but I am trying to understand the context.

lorban:
Sorry, but this code predates me by a long shot so I can't really say why it was introduced.
The few other places were `Os.isSolaris()` are called speak about working around bugs in JDK 1.4 so my best guess is that this was the reason for this one too. Except for a bug, I fail to see why the memory monitor would need a high priority. And AFAIK, all threads start at `Thread.MAX_PRIORITY` by default on Solaris and Linux.

jd0-sag:
This is good to merge but can you rebase it, first?  Some things have been shifting around.

PR #236, opened 2016-06-25
Issue 232 & 233

myronkscott:
Separate synchronization for connectionMade and clientStopped

Add notify for health checker stop

Add a test for issue 232 and 233


jd0-sag:
These changes improve things, so I will merge them.  My only concern is whether c439bb6562cec06a33a40c318d01e45256e8bc55 provably ends the thread before shutdown returns, or if it just narrows the window where it is still alive.

PR #241, opened 2016-06-29
workaround for solaris jvm that is unable to allocate more than 1.7GB…

lorban:
… of offheap when it's started with 2G or less heap

### DIFF
--- a/terracotta-kit/src/assemble/server/bin/start-tc-server.sh
+++ b/terracotta-kit/src/assemble/server/bin/start-tc-server.sh
@@ -83,7 +83,9 @@ setPluginClasspath;
 start=true
 while "$start"
 do
-${JAVA_COMMAND} -Xms2g -Xmx2g -XX:+HeapDumpOnOutOfMemoryError \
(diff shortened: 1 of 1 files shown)


myronkscott:
The server is way slimmed down from past iterations.  I suspect that 2G of heap is way overcommitting.  Can you please try using -server or -d64 flags and see if either of those fix the issue?

lorban:
@myronkscott : there seems to be a bug in the solaris JVM (and solaris only) that makes it throw OOME when a bit less than 2GB of offheap memory has already been allocated. The amount of free RAM on the machine doesn't matter, and as soon as you set the max heap size above 2GB (which is why I changed the default to 2049M) then everything works as it should.

I've reported the bug to Oracle, but since it's fairly easy and harmless to workaround, I though we might just do that now.

jd0-sag:
Ideally, we would use less heap than this for the server but if they have a bug which will limit our off-heap, this is at least a pretty harmless work-around until they fix it or it falls behind the supported versions.

PR #260, opened 2016-07-13
Fix #253 - Add client stracktrace to EntityExceptions

### LINKED ISSUE #253: EntityException on clients should report both local stack and server stack

ramsai1729:
Opening this PR for discussion


jd0-sag:
I think this justifies creating an issue against APIs.  We need to add a cause constructor to all of the entities and we might as well add accessors for the information within the specific exceptions (in the common case, things like class, name, and cause - in the other cases, the more detailed information).

ramsai1729:
will add some changes to this PR after completing Terracotta-OSS/terracotta-apis#127

PR #267, opened 2016-07-21
Diff lockless

myronkscott:
Fixes a number of issues along with a major refactoring of ManagedEntity interaction with ProcessTransactionHandler and replicated transaction handler.

Adds entity life cycling without server distributed locks.


tc-messaging/src/main/java/com/tc/entity/NetworkVoltronEntityMessageImpl.java
```diff
@@ -150,12 +159,21 @@ protected boolean hydrateValue(byte name) throws IOException {
...
+    }
+    
     return true;
   }
 
   @Override
   public EntityMessage getEntityMessage() {
-    // Network messages don't contain an identity message.
-    return null;
+    return this.message;
```
jd0-sag: So, does this mean that we are piggy-backing on that "inline message reference" idea used by IEntityMessenger and RetirementManager support?  I am surprised that we ended up finding  more general-case use for what, at the time, seemed like a bit of a hack.

tc-messaging/src/main/java/com/tc/entity/MessageCodecSupplier.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.tc.object.EntityID;
+import org.terracotta.entity.EntityMessage;
+import org.terracotta.entity.EntityResponse;
+import org.terracotta.entity.MessageCodec;
+
+/**
+ *
+ */
+public interface MessageCodecSupplier<M extends EntityMessage, R extends EntityResponse> {
```
jd0-sag: It looks like we aren't really able to capitalize on the generics whenever we are using this interface so it seems like we should use the EntityMessage and EntityResponse, directly, and rely on standard inheritance rules and down-casting.

tc-messaging/src/main/java/com/tc/entity/NetworkVoltronEntityMessageImpl.java
```diff
@@ -150,12 +159,21 @@ protected boolean hydrateValue(byte name) throws IOException {
...
     this.requiresReplication = getBooleanValue();
     this.oldestTransactionPending = new TransactionID(getLongValue());
+    
+    try {
+      if (this.type == Type.INVOKE_ACTION) {
+        MessageCodec<EntityMessage, EntityResponse> codec = supplier.getMessageCodec(this.entityDescriptor.getEntityID());
+        this.message = codec.decodeMessage(extendedData);
+      }
+    } catch (MessageCodecException exception) {
+//  swallow it
```
jd0-sag: You should flesh out more of the rationale behind why we can just swallow the exception.  I assume that the reason is that this is best-efforts, anyway, since the original use of the message field was a corner-case so we already have the machinery in place to attempt the decode, later on, if we don't run it here.

lockless-connection-impl/src/main/java/com/terracotta/connection/EntityClientServiceFactory.java
```diff
@@ -0,0 +1,69 @@
...
+ *  the specific language governing rights and limitations under the License.
+ *
+ *  The Covered Software is Terracotta Core.
+ *
+ *  The Initial Developer of the Covered Software is
+ *  Terracotta, Inc., a Software AG company
+ *
+ */
+
+/*
```
jd0-sag: License duplicated.

tc-messaging/src/main/java/com/tc/entity/MessageCodecSupplier.java
myronkscott: Makes sense

tc-messaging/src/main/java/com/tc/entity/NetworkVoltronEntityMessageImpl.java
myronkscott: That is right.  This is best efforts.  Another attempt is made to decode on the PTH were the client can be properly notified

myronkscott: I'll add commenting

tc-messaging/src/main/java/com/tc/entity/NetworkVoltronEntityMessageImpl.java
myronkscott: Yes.  This fits nicely here.

dso-l2/src/main/java/com/tc/l2/state/StateManagerImpl.java
```diff
@@ -210,7 +210,12 @@ private synchronized void moveToPassiveReady(Enrollment winningEnrollment) {
     logger.debug("moving to passive ready " + state + " " + winningEnrollment);
     if (state == START_STATE) {
       setActiveNodeID(winningEnrollment.getNodeID());
-      state = PASSIVE_UNINITIALIZED;
+      state = (startState == null) ? PASSIVE_UNINITIALIZED : startState;
+      if (!state.equals(PASSIVE_STANDBY) && !state.equals(PASSIVE_UNINITIALIZED)) {
+// TODO:  make sure this is the proper way to handle this.
```
ramsai1729: Don't we need to re-sync a server with Active if it went down as a PASSIVE_STANDBY?

ramsai1729: Do we even hit this code path in case of restartable server?
````
} else if (activeNode.isNull() || activeNode.equals(winningEnrollment.getNodeID())
               || clusterMsg.getType() == L2StateMessage.ELECTION_WON) {
      // There is no active server for this node or the other node just detected a failure of ACTIVE server and ran an
      // election and is sending the results. This can happen if this node for some reason is not able to detect that
      // the active is down but the other node did. Go with the new active.
      if (startState == null || startState == START_STATE) {
        moveToPassiveReady(winningEnrollment);
        if (clusterMsg.getType() == L2StateMessage.ELECTION_WON_ALREADY) {
          sendOKResponse(clusterMsg.messageFrom(), clusterMsg);
        }
      } else {
        //this server was started with persistent data, this server could be either a ACTIVE or PASSIVE before
        // it went down. In any case, don't start this server in passive state.
// TODO: send a negative response so the active zaps this server.  This is a bad way to agree.  find a better way
        if (clusterMsg.getType() == L2StateMessage.ELECTION_WON_ALREADY) {
          sendNGResponse(clusterMsg.messageFrom(), clusterMsg);
        }
      }
    } else {
````

File: dso-l2/src/main/java/com/tc/objectserver/handler/ReplicatedTransactionHandler.java [hunk 6]
```diff
@@ -235,61 +217,99 @@ private void replicatedMessageReceived(ReplicationMessage rep) throws EntityExce
...
+    try {
+      if (rep.getReplicationType() == ReplicationMessage.ReplicationType.INVOKE_ACTION) {
+        msg = entityManager.getMessageCodec(entityID).decodeMessage(extendedData);
+      }
+    } catch (MessageCodecException codec) {
+      throw new RuntimeException(codec);
+    }
+    payload = new MessagePayload(extendedData, msg, rep.getConcurrency());
     ReplicationMessage.ReplicationType replicationType = rep.getReplicationType();
     boolean didAlreadyHandle = false;
```
myronkscott: remove didAlreadyHandle.  It is no longer a factor

dso-l2/src/main/java/com/tc/objectserver/entity/EntityManagerImpl.java
```diff
@@ -98,15 +98,26 @@ public void enterActiveState() {
...
     // Valid entity versions start at 1.
     Assert.assertTrue(version > 0);
     ManagedEntity temp = new ManagedEntityImpl(id, version, noopLoopback, serviceRegistry.subRegistry(consumerID),
         clientEntityStateManager, this.eventCollector, processorPipeline, getVersionCheckedService(id, version), this.shouldCreateActiveEntities, canDelete);
-    if (entities.putIfAbsent(id, temp) != null) {
-      throw new EntityAlreadyExistsException(id.getClassName(), id.getEntityName());
+    ManagedEntity exists = entities.putIfAbsent(id, temp);
+    return exists != null ? exists : temp;
+  }
+
```
myronkscott: Check if unused.  Remove if so.

dso-l2/src/main/java/com/tc/l2/state/StateManagerImpl.java
myronkscott: If the server comes back as PASSIVE_STANDBY it is the responsibility of the active server to zap it if it is out of sync.

myronkscott: Yes.  we can enter this code path via handleElectionAbort.

PR #282, opened 2016-08-17
multi message return path

myronkscott:
Performance optimization for message batching on the return path.  Includes bug fixes to the messaging layer

### DIFF
--- a/common/src/main/java/com/tc/net/protocol/NetworkLayer.java
+++ b/common/src/main/java/com/tc/net/protocol/NetworkLayer.java
@@ -75,7 +75,7 @@ public interface NetworkLayer {
 
   public NetworkLayer getReceiveLayer();
 
-  public void send(TCNetworkMessage message);
+  public void send(TCNetworkMessage message) throws IOException;
 
   public void receive(TCByteBuffer[] msgData);
 
--- a/common/src/main/java/com/tc/net/protocol/delivery/OnceAndOnlyOnceProtocolNetworkLayerImpl.java
+++ b/common/src/main/java/com/tc/net/protocol/delivery/OnceAndOnlyOnceProtocolNetworkLayerImpl.java
@@ -418,7 +418,12 @@ public class OnceAndOnlyOnceProtocolNetworkLayerImpl extends AbstractMessageTran
     // this method doesn't do anything at the moment, but it is a good spot to plug in things you might want to do
     // every message flowing down from the layer (like logging for example)
     if (this.sendLayer.isConnected()) {
-      this.sendLayer.send(msg);
+      try {
+        this.sendLayer.send(msg);
+      } catch (IOException ioe) {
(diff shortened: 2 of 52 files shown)


jd0-sag:
I will keep my questions/comments high-level, out here.

Regarding 5d5baba6155e8d00fff78a38f34b80ca1eaede3f:

- I assume that it is safe to purge the stale TCMessageType instances since we will never be connected to a version of the system which still uses them
- if we want a "always last" message type, we should probably just explicitly make one.  Something unused, with a clear name, so we don't renumber the final message between minor releases and cause problems
-what is the main driver for this change?  Is it performance or just simplicity/size?

Regarding 2a69e2ec365438bfebf2c5f76edad35dc2c3660a:

- the IOException does allow problems to be realized at a higher-level than where they were being discarded, before, but there are still some places where we only log this.  Are those cases safe?  (should probably add a one-liner to those to explain why they can be dropped to a log, only)
- we probably want to come up with some kind of rule around how to manage synchronized entry-points, in `MessageTransportStatus`:  make the public/private of the methods explicit and make only one type synchronized (probably the public methods, given we are synchronizing on the instance, itself)
- should we be using `TCMessageType.NOOP_MESSAGE` in place of those now-removed lock messages or should those just be removed (also, weren't these the same messages removed in the previous commit?)

b2c3371b21c3dd25b2c003208bef7492b3668de6 seems straight-forward enough.


In general, I noticed that there is still some commented-out code which should just be deleted, instead of merged that way.  That is a trivial concern, though, just in case you happen to be going back into those commits, anyway.

Regarding 5d6205d32e573e9751f8a1cc2cf9add8c9f9c37a:  I am not sure why this is changing and I am worried that we are moving to an unordered collection given that I assumed we were relying on the ordered nature (although I would not be completely surprised if the algorithm has a different way of ensuring correctness).

Regarding 1bd1835e2a9f3410ec64cf86f20245fe60fab211:

- Is it possible for the client to fail to find the in-flight message on the receipt of various acks and responses?  Won't it only be re-sent if the client decided to re-send the message?  I wouldn't expect it to be missing or does this new approach potentially create a sort of "echo" as a side-effect of its implementation?
- do we not need to handle the error returns paths (the few there are) through this mechanism?
- am I correct in assuming that this only applies to INVOKE type messages? (if true, this may imply that we should look into further specializing this path, in the future)

myronkscott:
Regarding the logging of dropped messages.  Since this is preserving old behavior, I'm not actually sure what is safe and not safe so handling each case will only be double logging.

myronkscott:
Regarding the resent returns.  The problem is not reconnects to the same server as previously discussed but rather messages queued in the incoming message queue and hydrate stages on reconnect.  While it is possible to flush all the queues leading up to inflight messages before determining what to resend.  It may be more pragmatic to just document the fact that resent messages may result in multiple returns.

PR #283, opened 2016-08-17
Update dependencies and add monitoring tests

jd0-sag:
Updated dependencies:
-terracotta-apis.version - 1.0.4.beta
-terracotta-configuration.version - 10.0.4.beta

Also added some tests regarding the new M&M Serializable behavior


jd0-sag:
Self-merging since this is just a version bump and a new unit test.

PR #287, opened 2016-08-25
Missed update of config version:

jd0-sag:
-terracotta-configuration 10.0.5.beta


jd0-sag:
Will self-merge as this is just a dependency update.

PR #294, opened 2016-08-26
Populate consumerID in PlatformEntity

jd0-sag:
-updated dependencies for terracotta-apis 1.0.6.beta and terracotta-configuration 10.0.6.beta
-this is mostly just plumbing the consumerID down through ManagementTopologyEventCollector


jd0-sag:
Self-merging to get the release out, as this change was discussed, earlier.

PR #311, opened 2016-09-20
[DO NOT MERGE] Message typing for replication

myronkscott:
This is a follow along for terracottas-apis#154 PR to show implementation of both methods.

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/entity/AbstractServerEntityRequestResponse.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/entity/AbstractServerEntityRequestResponse.java
@@ -31,8 +31,6 @@ import com.tc.objectserver.api.ServerEntityAction;
 import com.tc.objectserver.api.ServerEntityRequest;
 import com.tc.objectserver.api.ServerEntityResponse;
 import com.tc.util.Assert;
-import java.io.IOException;
-import java.util.Collections;
 
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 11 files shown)


jd0-sag:
I suppose that the overall question regarding all of this is "what problem is this trying to solve?"  Is this just another way of passing a replication flag or does it give platform or entities other benefits?

I suppose that one interesting thing we get out of this is that it allows "passive-only" messages, which is something entities have been interested in but we have never considered making first-class concepts within the system.

I guess the main question is what advantages the entities can get out of this beyond what we could get via flags in the InvocationBuilder.  If it provides some benefit or clarity for them, we should probably go forward with one of these.

Personally, I prefer the annotation since it avoids explicit type-checks, is less rigidly tied to inheritance patterns, and _seems_ to more directly map to what we are saying (this message "has these attributes" as opposed to "is these things" - although that is arguably personal preference).

Even if we proceed with neither of these, the "non-replicated message" (that is, replicate as NOOP) that you have here should be pulled in as something we actually want to use (since it is currently missing).

myronkscott:
The problem with the invocation builder method is that replication is inherent to the message type not the invocation of that message.  It's too easy to write...
```
endpoint.beginInvoke().message(new MutationMessage()).replicate(true).invoke();
endpoint.beginInvoke().message(new MutationMessage()).replicate(false).invoke();
```
Which in my mind would be a bug.  By tying execution the the message type, it is more clear where a message type is supposed to run.

The notion of the passive only message would help the case where InterEntityMessenger would do an invoke with deferred retirement and we want the sub-message to be ignored on the active.

PR #321, opened 2016-10-13
Follow terracotta api changes for PlatformConfiguration

ramsai1729:


jd0-sag:
You will need to revert the pom.xml change where the version dependency is set but this will need to be worked into a release, since it is a change in the APIs.

PR #328, opened 2016-10-20
interrupt stuck threads

myronkscott:
TCGroupManagerNodeJoinedTest may have stuck threads that need interrupting


myronkscott:
test this please

PR #368, opened 2016-11-25
Removed all references to IPersistentStorage

jd0-sag:
-this removes the now-deprecated FlatFileStorage implementation and the emulated IPersistentStorage implementation


jd0-sag:
(self-merging, as discussed)

PR #394, opened 2016-12-10
Improve lifecycle and sync interaction

myronkscott:
Lifecycle and sync interaction is improved by not blocking the process transaction handler

Small improvement to not wait for message completion on passives with execution was designated to be only active.

debug message logging

fix an improper assertion about start the start of a replicated message stream.

### DIFF
--- a/connection-impl/src/main/java/com/terracotta/connection/entity/TerracottaEntityRef.java
+++ b/connection-impl/src/main/java/com/terracotta/connection/entity/TerracottaEntityRef.java
@@ -162,6 +162,8 @@ public class TerracottaEntityRef<T extends Entity, C> implements EntityRef<T, C>
     } catch (EntityException e) {
       if (e instanceof EntityNotFoundException) {
         throw (EntityNotFoundException)e;
+      } else if (e instanceof PermanentEntityException) {
+        throw (PermanentEntityException)e;
       } else {
         throw new RuntimeException("unexpected", e);
       }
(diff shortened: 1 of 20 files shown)


connection-impl/src/main/java/com/terracotta/connection/entity/TerracottaEntityRef.java
```diff
@@ -107,17 +110,42 @@ public String getName() {
...
+          entityManager.createEntity(entityID, version, entityClientService.serializeConfiguration(configuration)).get();
+          return null;
+        }
+      });
     } catch (EntityException e) {
       // Note that we must externally only present the specific exception types we were expecting.  Thus, we need to check
       // that this is one of those supported types, asserting that there was an unexpected wire inconsistency, otherwise.
       e = ExceptionUtils.addLocalStackTraceToEntityException(e);
-      if (e instanceof EntityNotProvidedException) {
+      if (e instanceof EntityBusyException) {
```
jd0-sag: Can this exception be encountered here?

jd0-sag:
I will need you to walk me through some of this cooperative scheduling of lifecycle operations against sync since I can't shake a feeling that we are going to return success in some cases where we need to wait and won't be able to tell.

A conversation and maybe some more comments and/or tests might set my mind at ease.

connection-impl/src/main/java/com/terracotta/connection/entity/TerracottaEntityRef.java
myronkscott: I pushed everything a layer lower.  I probably just missed this

PR #406, opened 2016-12-19
Misc Refactorings + bug fixes

myronkscott:


dso-l2/src/main/java/com/tc/objectserver/entity/MessageDecoder.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package com.tc.objectserver.entity;
+
+import org.terracotta.entity.EntityMessage;
+import org.terracotta.entity.MessageCodecException;
+
+/**
+ *
+ */
+public interface MessageDecoder {
```
jd0-sag: I don't see what this level of indirection buys us or how it is being implemented.  It looks like it depends on an API change.

myronkscott: MessagePayload needs to be able to decode both SyncMessages (SyncMessageCodec) and EntityMessages (MessageCodec).  MessageDecoder is used to handle both via lambda expression.  The main point of this commit is to pull the codecs from the ManagedEntity instead of the EntityServerService

dso-l2/src/main/java/com/tc/objectserver/handler/ReplicationSender.java
```diff
@@ -286,12 +269,16 @@ public void validateSending(ReplicationMessage msg) {
...
+      return messageId++;
+    }
+    
     public boolean isComplete() {
       return lastSent == SyncReplicationActivity.ActivityType.SYNC_END;
     }
     
     private SyncReplicationActivity.ActivityType validate(SyncReplicationActivity.ActivityType type, SyncReplicationActivity.ActivityType compare) {
-      switch (type) {
+      /*switch (type) {
```
myronkscott: I turned it off temporarily.  Will turn it back on.  Good catch.

PR #413, opened 2016-12-21
[DO NOT MERGE] Reorder Activation of the LocalMonitoringProducer

myronkscott:
Reorder activation of LocalMonitoringProducer to have the rebuild event populated before the reconnect logic takes over.


common/src/main/java/com/tc/async/impl/StageController.java
```diff
@@ -69,10 +84,14 @@ public void transition(ConfigurationContext cxt, com.tc.util.State old, com.tc.u
       }
     }
     for (String s : coming) {
-      Stage<?> st = cxt.getStage(s, Object.class);
-      if (st != null && !leaving.contains(s)) {
-        st.start(cxt);
-        st.unpause();
+      if (s.startsWith("TRIGGER-")) {
```
jd0-sag: Should we do something like use 2 different collections for this instead of checking the string?

myronkscott: I'm not sure how you would do that.  Order is important.  Other option would be to make it a Set<Object> and type sniff Runnable and Strings

PR #421, opened 2016-12-26
Issue 382

jd0-sag:
This continues some refactoring around the ReplicationMessage and completes the conversion of it into a container type for buffering the transmission of SyncReplicationActivities to passives.


jd0-sag:
If you are happy with these changes, feel free to merge.  Ideally, we can crack a follow-on item to coalesce the buffering mechanisms and add more unit tests specific to how they work.

In terms of my local testing, this change produced no real performance change.  Upon investigation, it is because very few of the message batches contained more than 1 activity.  If I deliberately added a stall of 100 ms to the sent callback, the batches grew to sometimes hundreds of activities.  All the tests still passed so the meaning/order of the activities was unaltered.

If you have any questions/concerns, we can discuss on Wednesday.

myronkscott:
Was your test single or multi-machine?

PR #428, opened 2016-12-30
Fix #427 - capture any fetches prior to getting the connected call.  

### LINKED ISSUE #427: addNode() for fetch comes after addNode() for client connection
This is a simple use case, 1 active server. The issue is random (test sometimes fail). The addNode() call that setup the tree at client connection is done after the addNode() call that inform us about a fetch.

```
2016-12-29 12:44:00,173 [WorkerThread(request_processor_stage, 3, 4)] TRACE org.terracotta.management.service.monitoring.IStripeMonitoringPlatformListenerAdapter - [0] addNode(testServer0, [platform, fetched], 2_org.terracotta.management.entity.management.client.ManagementAgentEntityManagementAgent_1)
2016-12-29 12:44:00,180 [WorkerThread(client_handshake_stage, 0)] TRACE org.terracotta.management.service.monitoring.IStripeMonitoringPlatformListenerAdapter - [0] addNode(testServer0, [platform, clients], 2)
```

So obviously we fail later.

Here is the server log.

[stdout.log.txt](https://github.com/Terracotta-OSS/terracotta-core/files/677619/stdout.log.txt)

FYI @anthonydahanne : the test `TopologyIT.notifications_have_a_source_context` randomly fails because of that.

myronkscott:
This can happen when the handshake thread gets descheduled in exactly the right place.  Ordering the calls in the top collector is the best way to handle this.

### DIFF
--- a/common/src/test/java/com/tc/net/core/TCWorkerCommManagerTest.java
+++ b/common/src/test/java/com/tc/net/core/TCWorkerCommManagerTest.java
@@ -73,6 +73,12 @@ public class TCWorkerCommManagerTest extends TCTestCase {
   TCLogger          logger               = TCLogging.getLogger(TCWorkerCommManager.class);
   List<ClientMessageTransport> transports = new ArrayList<ClientMessageTransport>();
 
+  public TCWorkerCommManagerTest() {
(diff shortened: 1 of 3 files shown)


jd0-sag:
This is based oddly - should be rebased before merge.

PR #442, opened 2017-01-12
Client comms rework

myronkscott:
Add dynamic redirect from passive server to client
Add diagnostic port to get basic state of a server not yet connected to the cluster
supporting interface changes and relaying


common/src/main/java/com/tc/net/protocol/transport/ClientMessageTransport.java
```diff
@@ -112,30 +115,72 @@ public NetworkStackID open() throws TCTimeoutException, IOException, MaxConnecti
...
+  TCConnection connect(TCSocketAddress sa) throws TCTimeoutException, IOException {
+    TCConnection connection = this.connectionManager.createConnection(getProtocolAdapter());
+    fireTransportConnectAttemptEvent();
+    try {
+      connection.connect(sa, this.timeout);
+    } catch (IOException e) {
+      connection.close(100);
+      throw e;
+    } catch (TCTimeoutException e) {
+      throw e;
```
lorban: Why is there no `connection.close(100)` here?

lorban:
I went through all those changes, and basically you've "just" modified some of the plumbing to be able to send a redirect to a client.

I didn't see anything obviously wrong, but the comms layer's FSM is so complicated that nothing is obvious. These changes do make me worry despite the fact that I can't see anything wrong.

common/src/main/java/com/tc/net/protocol/transport/ClientMessageTransport.java
myronkscott: It looks like TCTimeout causes the underlying connection to do its own cleanup but adding the close for safe measure seems like a good idea.

myronkscott:
Ludovic.  Thanks for going through the PR.  Yes.  This change has me very concerned if for no other reason, the shear size of the diff.  I've tried to push the comms stack into a more manageable shape.  Time will tell if I succeeded.

Issue #439, opened 2017-01-10
Entity order in passive sync and restart must be well-defined


jd0-sag:
Currently, when synchronizing entities to a new passive or reloading entity state, the order in which the entities are instantiated is not well-defined.

This causes problems if the entities assume a relative order of creation (which a client can ensure for them upon initial creation).

It should be possible to ensure that the creation order is the same, when synchronized or restarted, as when initially created.

For restart, this is relatively straight-forward as the reload of state happens at a specific point in time, before any messages are re-sent.

For passive synchronization, this is more complex since the sync happens _while_ new messages are actively arriving.  This means that our passive sync state machine needs to change:  instead of synchronizing each entity, one at a time, we will probably need to pass all entity creation data in the first message to start the sync.  At that point, all the entities will be instantiated and created, in order.  From then on, the sync can proceed with moving their data, in-order (much as is currently done).

Destroy and reconfigure calls will also need some attention, here, since they will be applied to something which exists on the passive but was never synchronized, at all.

jd0-sag:
@ljacomet Add any other concerns you had regarding this matter, here.

jd0-sag:
Note that the order of reload is also currently undefined for restart cases since the `EntityPersistor` stores the current entities as a map, not a list.  That said, this data could be sorted by the caller, since the returned data contains the `consumerID`.  The `consumerID` is currently a monotonic counter, per-server, so a later a `consumerID` implies an entity which was created later.

jd0-sag:
Another question we need to answer is if this ordering needs to apply to promotion of entities, as well (replacing a passive with an active entity when and existing passive becomes the new active).

I suspect that we need that to be done in order, as well.

PR #443, opened 2017-01-12
Re #439: Sort entities by consumerID before promotion or restart

jd0-sag:
-we want to promote and restart entities in the same order they were originally instantiated
-note that the passive sync case will be handled in a later commit

Ideally, this would merge after the other PR regarding diagnostic port support.

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/entity/EntityManagerImpl.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/entity/EntityManagerImpl.java
@@ -37,6 +37,9 @@ import com.tc.text.PrettyPrinter;
 import com.tc.util.Assert;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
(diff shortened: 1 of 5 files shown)


dso-l2/src/main/java/com/tc/objectserver/entity/EntityManagerImpl.java
```diff
@@ -103,7 +106,20 @@ public void enterActiveState() {
...
+        @Override
+        public int compare(ManagedEntity o1, ManagedEntity o2) {
+          long firstID = o1.getConsumerID();
+          long secondID = o2.getConsumerID();
+          return (firstID > secondID)
+              ? 1
+              : (firstID < secondID)
+                ? -1
+                : 0;
+        }});
```
myronkscott: Assert on equals here?

jd0-sag: As in, assert that the values are NOT equal?  I suppose that would make sense since each of these should be unique.

PR #450, opened 2017-01-19
Fix #449: Expose platform restartability info

### LINKED ISSUE #449: Expose platform restartability info
This info is useful to entities to handle cases like creating restartable entities on non-restartable platform etc.

ramsai1729:
- Expose platform restartability info via PlatformConfiguration
- Some refactorings in TerracottaServiceProviderRegistryImpl to expose this information via PlatformConfiguration


myronkscott:
Will discuss later

Issue #466, opened 2017-01-31
Entity throwing in createNew are still tracked as existing


ljacomet:
As in terracotta-oss/tc-passthrough-testing#88, the recent changes introduced by #464 are incomplete.
If an entity throws during createNew it is not removed from the tracked active or passive entities.
This prevents any further successful creation of the same entity.

myronkscott:
fixed by #467

PR #467, opened 2017-01-31
Issue #466 Do not track entity unless createNew succeeds

ljacomet:


ljacomet:
Tested with Ehcache that this fixes the issues encountered in tests. However did not really figure out how to write a test asserting the same in the PR.

PR #469, opened 2017-01-31
Fix #468 - Check for ManagedEntity removal on every flush

### LINKED ISSUE #468: ManagedEntities need to be GC'd
ManagedEntities that are removable need to be properly GC'd from the EntityManager

myronkscott:
Be more aggressive about GC of removable ManagedEntities


jd0-sag:
The first 2 commits might be clearer if they were merged.  What do you think?

The 3rd commit comment should be re-worded so that GitHub will do the auto-close (the space is in the wrong spot for it to recognize the reference).

While I would slightly prefer `LocalPipelineFlushMessage` to be split for this purpose, I have no strong feeling regarding that (probably not worth changing unless you think that is a good idea).

PR #474, opened 2017-02-01
Improve handling of IPlatformPersistence.fastStoreSequence() future

ramsai1729:


dso-l2/src/main/java/com/tc/objectserver/handler/ReplicatedTransactionHandler.java
```diff
@@ -282,17 +285,21 @@ private void replicatedActivityReceived(ServerID activeSender, SyncReplicationAc
 
     // Send the RECEIVED ack before we run this.
     ackReceived(activeSender, activity);
```
jd0-sag: I think that the order is wrong here:  We need to wait on the future before sending the ackReceived, not the ack for complete.

Issue #465, opened 2017-01-31
Identify slow points in invocation path


jd0-sag:
We need to improve the overall performance of the system (especially in the case of multi-server stripes).  This requires measuring the current cost of these operations so we have some guidance on where to drill down further.

For these initial numbers, I created a single-client test which sends a sequence of 10000 invokes which do effectively nothing (but manually requesting replication - implicit replication disabled), measuring the average time it takes from before the send to when the `RECEIVED` ack returns to when the `COMPLETED` or `RETIRED` ack returns (did 10000 runs of each, interleaved to avoid skew due to trends over time).  This was then tested on a single-server stripe versus a 2-server stripe.  NOTE:  Client and server(s) run on the same machine (avoids network overhead and this test isn't CPU heavy, just a measure of total path length).

Each run produced an average of the various ack timings for all configurations.  I ran 3 sets and averaged the corresponding results (there was relatively little variability between them).

Single-server:

* `RECEIVED` to `COMPLETED`: 25% longer
* `RECEIVED` to `RETIRED`: 54% longer

2-server:

*`RECEIVED` to `COMPLETED`: 77% longer
*`RECEIVED` to `RETIRED`: 177% longer

Overall cost of adding a passive:

*`RECEIVED` when waiting for `COMPLETED`: 9% longer
*`RECEIVED` when waiting for `RETIRED`: 13% longer
*`COMPLETED`: 55% longer
*`RETIRED`: 104% longer

Overall cost of waiting for `RETIRED` (above just waiting for `COMPLETED`):

* versus single-server `RECEIVED`: 5%
* versus single-server `COMPLETED`: 30%
* versus 2-server `RECEIVED`: 9%
* versus 2-server `COMPLETED`: 70%

Oddities:

* In all runs, waiting for `RETIRED` instead of `COMPLETED` made the time until `RECEIVED` arrived take longer, which wasn't expected.
* If the tests aren't interleaved, the set which runs later performs far better.  This remained true even after introducing a "warming" run, before the collection runs.  I had expected this warming to minimize such an effect but this wasn't true.

Consistencies with expectations:

* Waiting for `RETIRED`, on the passive, effectively doubles the path-length (since the manual replication means the passive can only start after the active has mostly run the message) so the 70% overhead is believable.  Single-server cost of 30% is believable since it only adds another invoke, not another network round-trip.
[...]

PR #475, opened 2017-02-02
Perf testing and discussion

myronkscott:
After conversation it was discovered by @jd0-sag that creating replicated messages was unreasonably expensive.  See #465.  By generating load internally on the server, two hotspots were identified.  The first was the switch case statement in RequestProcessor. createReplicationActivity as a possible culprit.  Changed to an enum map and profiling data suggests that it has helped.  The debug statement optimization is probably trivial.

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
@@ -270,7 +270,12 @@ public class ManagedEntityImpl implements ManagedEntity {
     }
     
     SchedulingRunnable next = new SchedulingRunnable(desc, request, payload, r, ckey);
-    results.setWaitFor(next);
+    
+    if (isActive()) {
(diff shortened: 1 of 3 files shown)


myronkscott:
Happy to remove the load generator commit.  It is useful but questionable to live in the non-test codebase.  Requesting comment

jd0-sag:
I don't quite understand what these changes are doing or how they address some of those issues found in the investigation:

* The comment in b5ccfc3c6eb098b5f0f6adcf3c4c4fa08660ff16 doesn't seem related to what is happening in the change.
* I am a little surprised that 6e4e8f0c0bb6e8709b8dc0d557e93354ef70916a makes things better (especially when this wasn't really a hot point) but I could imagine an EnumMap maybe being faster than a switch (since it will generate a sequence of compares instead of a sort of "index by value").  The change does make things into a more declarative mapping, which I do like better than the old imperative structure.
* c745b8c8ebd32c9ffcb28f064a786b7ef7ea31de makes sense although I sometimes wonder if we should just remove that debugId, altogether.

PR #476, opened 2017-02-02
Fix #471: Remove receivedTransactionCount persistence

### LINKED ISSUE #471: TransactionOrderPersistor: consider removing receivedTransactionCount persistence
This information can be inferred from transaction information associated with existing clients

ramsai1729:


jd0-sag:
I was wondering if we do need to store this when a client disconnects but I would like to think that we can just proceed with this change since it is simple and will work, in the general case.

@myronkscott Do you think that there are any cases where this could cause an issue (electing the wrong leader, really) or have we convinced ourselves that that could only happen in the cases where all options were equivalent?  I _suspect_ that we are good here but I just wanted to float the idea.

jd0-sag:
We have convinced ourselves that the only case where this could matter, which we hadn't previous discussed (multiple restartable passives, one newly joining and thinking it can become leader), are already mitigated by more fundamental logic within the election code.

This change looks good.

PR #477, opened 2017-02-06
Rollup of Messaging Changes

myronkscott:
Includes:
Changes EntityDescriptor definitions.  Trying to follow unix file system semantics.  There are now 3 types of descriptors.  Create/Destroy with entityid and version, Fetch/Release with fetchid (long) and ClientInstanceID, Invoke with fetchID only

Wiring requested ACKs to the server so optimization decisions can be made there.

Minor bug fixes.


dso-l1/src/main/java/com/tc/object/EntityClientEndpointImpl.java
```diff
@@ -50,18 +52,35 @@
   /**
    * @param entityDescriptor The server-side entity and corresponding client-side instance ID.
    * @param invocationHandler Called to handle "invokeAction" requests made on this end-point.
+   * @param fetchID
```
jd0-sag: It looks like this was removed.  The JavaDoc should probably be updated, anyway.

There is other JavaDoc which is rendered stale by this change but I don't want to block this based on those.

dso-l1/src/main/java/com/tc/object/ClientEntityManagerImpl.java
```diff
@@ -426,12 +426,17 @@ private void throwClosedExceptionOnMessage(InFlightMessage msg) {
...
-      byte[] config = internalRetrieve(fetchDescriptor);
+      byte[] raw = internalRetrieve(fetchDescriptor);
+      ByteBuffer br = ByteBuffer.wrap(raw);
+      final long fetchID = br.getLong();
+      byte[] config = new byte[br.remaining()];
+      br.get(config);
       // We can only fail to get the config if we threw an exception.
-      Assert.assertTrue(null != config);
+      Assert.assertTrue(null != raw);
+      final
```
jd0-sag: This probably isn't supposed to be here.

dso-l2/src/main/java/com/tc/objectserver/entity/EntityManagerImpl.java
```diff
@@ -189,21 +195,33 @@ public boolean removeDestroyed(EntityID id) {
...
-  public Optional<ManagedEntity> getEntity(EntityID id, long version) throws EntityException {
+  public Optional<ManagedEntity> getEntity(EntityDescriptor descriptor) throws EntityException {
+    if (descriptor.isIndexed()) {
+      return getEntity(descriptor.getFetchID());
+    } else {
+      return getEntity(descriptor.getEntityID(), descriptor.getClientSideVersion());
+    }
+  }
+  
+  private Optional<ManagedEntity> getEntity(long idx) {
```
jd0-sag: Can these names be made more descriptive?  They are currently both called "getEntity" but interpret their arguments very differently.

dso-l2/src/main/java/com/tc/objectserver/entity/EntityManagerImpl.java
```diff
@@ -245,7 +263,7 @@ public boolean removeDestroyed(EntityID id) {
       Assert.assertNotNull(service);
       Object oldService = entityServices.putIfAbsent(typeName, service);
       // This needs to be null or else there was some kind of unexpected concurrent access which would have caused failure or a duplicate entry.
-      Assert.assertNull(oldService);
+      Assert.assertNull(oldService);
```
jd0-sag: There are a few extraneous whitespace changes in this commit, such as this, which should probably be kept from cluttering the version history.

dso-l2/src/main/java/com/tc/objectserver/entity/EntityManagerImpl.java
```diff
@@ -57,6 +58,7 @@
 public class EntityManagerImpl implements EntityManager {
   private static final TCLogger LOGGER = TCLogging.getLogger(EntityManagerImpl.class);
   private final ConcurrentMap<EntityID, ManagedEntity> entities = new ConcurrentHashMap<>();
+  private final ConcurrentMap<Long, ManagedEntity> entityIndex = new ConcurrentHashMap<>();
```
jd0-sag: Since this just uses a Long as the key, you should add a comment here to describe what the key actually means.

dso-l2/src/main/java/com/tc/objectserver/handler/ReplicatedTransactionHandler.java
```diff
@@ -278,7 +276,7 @@ private void replicatedActivityReceived(ServerID activeSender, SyncReplicationAc
     ClientID sourceNodeID = activity.getSource();
     TransactionID transactionID = activity.getTransactionID();
     TransactionID oldestTransactionOnClient = activity.getOldestTransactionOnClient();
-    EntityDescriptor descriptor = activity.getEntityDescriptor();
+//    EntityDescriptor descriptor = activity.getEntityDescriptor();
```
jd0-sag: This line should just be deleted.

dso-l2/src/test/java/com/tc/objectserver/entity/ClientEntityStateManagerImplTest.java
```diff
@@ -44,7 +44,6 @@
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
-import static com.tc.util.Assert.assertEquals;
```
jd0-sag: This is the only change in this file, probably not worth it.

myronkscott:
I will convert fetchID to a node type.  We do this a lot of places.  I'm not totally pleased with the pattern but it will make map usage nicer at least.

Will see about EntityDescriptor 

Thought about just making received like replicated or even replacing replicated (redundant due to ExecutionStrategy) but I figured enough was up in the air that more information was better than less.

I had the same concern about received sending on the passive but I think everything is relative and since the passive is doing very little (given concurrency index, etc) and because of batching, I think this matters very little in the overall throughput.  The more we can push to the RequestProessor threads the better.

Will look at whitespace problem.

Thanks.

dso-l2/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
```diff
@@ -1040,7 +1047,7 @@ private void end() {
   //  queue so if it is not empty, push a noop.  
         flushLocalPipeline.completed(id, fetchID, request.getAction());
       }
-    }
+    }
```
myronkscott: Fix whitespace

PR #480, opened 2017-02-09
Shorten the cluster 'name' by using the class simple name

chrisdennis:


galvan-support/src/main/java/org/terracotta/testing/rules/BasicExternalCluster.java
```diff
@@ -108,7 +108,19 @@ public BasicExternalCluster(File clusterDirectory, int stripeSize, List<File> se
 
   @Override
   public Statement apply(Statement base, Description description) {
```
gibsong: Did u test this?  I was pretty sure the issue was actually at line BasicExternalCluster:111 when it calls
File testParentDirectory = File.createTempFile(this.displayName, "", this.clusterDirectory);
This is the code that creates a directory name with the name of the fully qualified name followed by a random number.

chrisdennis: Right... this PR changes the content of the displayName string so that those paths are a little shorter.  Eg:
```
[cdennis@Chriss-MacBook-Pro galvan-support (bec-display-name *)]$ ls target/cluster/
BasicExternalClusterActivePassiveIT3211422405588413478						terminateAllServers(BasicExternalClusterImmediateControlIT)4674232810595424681
BasicExternalClusterClassRuleIT5121380554379762459						terminateOnePassive(BasicExternalClusterImmediateControlIT)4622088741602320130
CorruptConfigWithClassRuleIT3055325380871200805							testConnectionViaURI(BasicExternalClusterMethodRuleIT)123431684609481913
InvalidPathTestIT4645686398243518311								testDirectConnection(BasicExternalClusterMethodRuleIT)3553423029669824777
startAllServers(BasicExternalClusterImmediateControlIT)759001885528126494			waitForActive(BasicExternalClusterImmediateControlIT)113079066240505490
startOneServer(BasicExternalClusterImmediateControlIT)8362287716628713182			waitForRunningPassivesInStandby(BasicExternalClusterImmediateControlIT)5303871664680397283
terminateActive(BasicExternalClusterImmediateControlIT)3322173363172794770
```

gibsong: Unfortunately on my machine I have to drastically shorten the directory name.  This is the longest directory name that works for me "1234567890123456"  And this is the resulting path to the bat file:
C:\Users\ggib\dev\irepo\terracotta-enterprise\management\tms-system-tests\target\galvan\1234567890123456\testServer0\server\bin\start-tc-server.bat

jd0-sag:
How much of this problem can we solve, directly, before we have to say "Windows doesn't really work - just check things out closer to a root - err 'letter'"?

We can avoid aggravating the problem but, at some point, we have to acknowledge that this is a bug in something else.

jd0-sag:
I will merge this as it reduces the aggravation of the issue but, if we want a more complete solution, we probably need to root the tests externally to the build target.

PR #497, opened 2017-02-17
Wait for invoke action transaction order persistence futures during m…

ramsai1729:
…utli send

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/handler/ProcessTransactionHandler.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/handler/ProcessTransactionHandler.java
@@ -62,6 +62,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.function.Consumer;
 import java.util.concurrent.Future;
 import java.util.function.Predicate;
@@ -90,6 +91,7 @@ public class ProcessTransactionHandler implements ReconnectListener {
   
   private Sink<TCMessage> multiSend;
   private ConcurrentHashMap<ClientID, TCMessage> invokeReturn = new ConcurrentHashMap<>();
+  private ConcurrentHashMap<TransactionID, Future<Void>> transactionOrderPersistenceFutures = new ConcurrentHashMap<>();
   
   private void sendMultiResponse(VoltronEntityMultiResponse response) {
     multiSend.addSingleThreaded(response);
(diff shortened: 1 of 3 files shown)


ramsai1729:
test this please

jd0-sag:
Are there some tests which demonstrate that this is a win?  I just want to make sure that we can prove this is a benefit since it might make other attempts to move the received path (handle RECEIVED acks and transaction order write-back) out-of-line more complicated.

dso-l2/src/main/java/com/tc/objectserver/handler/ProcessTransactionHandler.java
```diff
@@ -106,6 +108,12 @@ public synchronized void reconnectComplete() {
     public void handleEvent(TCMessage context) throws EventHandlerException {
       NodeID destinationID = context.getDestinationNodeID();
       invokeReturn.remove((ClientID)destinationID, context);
+      if(context instanceof VoltronEntityMultiResponse) {
```
jd0-sag: Do we still need to wait for this, even it if isn't a multi-response, or are those transactions waited for elsewhere?  Minimally, we need a comment explaining why this is a special-case.

ramsai1729: AFAIK there are two types of messages getting added into multiSend stage `VoltronEntityMultiResponse` and `VoltronEntityAppliedResponse`, we are handling other type (`VoltronEntityAppliedResponse`) at line no `323`. Correct me if I missed anything.

jd0-sag: That is only for the failed path, though.  Additionally, what about non-invoke responses?  Do they all go through the multi-response path?  (while I would assume that they should, I seem to remember there being a reason why they didn't).

Additionally, can that call be moved from line 323 to here?  As it stands, we have a hidden assumption that these 2 paths are equivalent, even though they are in very different parts of the code, are run in different threads, and are not documented as such.

ramsai1729:
I have done some manual testing, results are better with batching implementation.

dso-l2/src/main/java/com/tc/objectserver/handler/ProcessTransactionHandler.java
ramsai1729: Life cycle requests responses are not batched and they are being handled using AbstractServerEntityRequestResponse.received()

ramsai1729: will check and move VoltronEntityAppliedResponse handling here.

ramsai1729: moved VoltronEntityAppliedResponse handling code here.

dso-l2/src/main/java/com/tc/objectserver/handler/ProcessTransactionHandler.java
```diff
@@ -106,6 +108,14 @@ public synchronized void reconnectComplete() {
...
+        VoltronEntityMultiResponse voltronEntityMultiResponse = (com.tc.entity.VoltronEntityMultiResponse) context;
+        for (TransactionID transactionID : voltronEntityMultiResponse.getReceivedTransactions()) {
+          waitForTransactionOrderPersistenceFuture(transactionID);
+        }
+      } else if(context instanceof VoltronEntityAppliedResponse) {
```
jd0-sag: This is definitely easier to follow.

Are these the only 2 types of messages which come in here?

If so, either make this a generic else (and document that there are only 2 types, prior to the cast) or add a generic else to document that there are only 2 types and put an assert there.

If not, add a generic else with some comments explaining why you can get there and why not handling the persistence future makes sense, in that case.

As it stands, I am worried that a change to the type will mean we fall out of this, incorrectly.

ramsai1729: I can see only 2 types from the code, so have added a assert to make sure future messages additions gets handled properly.

PR #500, opened 2017-02-22
Fix #499: VoltronEntityMultiResponseImpl: fix a corner case while wai…

### LINKED ISSUE #499: VoltronEntityMultiResponseImpl: fix a corner case while waiting on future.get()

ramsai1729:
…ting on future.get()


jd0-sag:
What did this `isSealed()` do?  That makes it sound like it should have been addressing this issue.

Otherwise, I assume that what you are doing is making sure that we don't keep adding to a response after someone has started looking at its contents.

I suppose that the call-sites of those methods which depend on this flag are already correctly switching to the new batch in these failure situations?

jd0-sag:
Talked to Myron about `isSealed()` and that turns out to be related to serializing the message, which is a much later operation.  Your change makes sense and I think that the recent changes to put the persistence wait point here are probably why the timing hole is more easily observed, now.

PR #501, opened 2017-02-22
Lifecycle journaling broken 

myronkscott:
On passives, since client source information was not propagated, lifecycle operations are not journaled.  As a result, resent lifecycle operations can give the wrong answer.

TODO: add testing.


jd0-sag:
Is this to address that problem we were talking about, yesterday evening?  Does this imply that the duplicated entity instance in the the state dump was a red herring or can this actually cause that to happen?

PR #505, opened 2017-03-03
Improve logging and address start-up race

jd0-sag:
This improves some of our logging but primarily introduces a proposal for how to address the rare start-up race observed in some restartable configurations when the newly-restarted server observes the last messages of an election before it is actually up and running to the point where it can handle such information.

d60f377 is the most relevant change and is probably worth discussing in case we want to address this a different way.


dso-l2/src/main/java/com/tc/l2/state/StateManagerImpl.java
```diff
@@ -141,7 +136,13 @@ public void startElection() {
     } else if (state == START_STATE || state == PASSIVE_STANDBY) {
       runElection();
     } else {
-      info("Ignoring Election request since not in right state");
+      info("Ignoring Election request since not in right state: " + this.state);
+    }
+    // Now that we are ready, put us into the started state and notify anyone stuck waiting.
+    synchronized (this) {
+      Assert.assertFalse(this.didStartElection);
+      this.didStartElection = true;
```
myronkscott: doesn't this need to go before the runElection() call?  If the thread stalls after runElection(), I think the reply could be missed.

myronkscott: nevermind.  I guess consumers of the result will wait for this to happen.

PR #507, opened 2017-03-08
(DO NOT MERGE) Issue 506

jd0-sag:
NOTE:  This is just to discuss if we want to address this problem, this way.  We should still try out some smaller solutions, if they work, since we will likely need to put this into a maintenance branch.  Given that this solution is also just a tactical work-around for the bind issue we found, it shouldn't be considered the strategic solution.  A smaller tactical solution might be better.


jd0-sag:
Closing since this was resolved by 55ee59a4416a21fbf475791a9992e95273366f06

PR #508, opened 2017-03-09
Retry l1 port bind

myronkscott:
For discussion

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
@@ -252,8 +252,10 @@ import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.io.Writer;
 import java.lang.management.ManagementFactory;
+import java.net.BindException;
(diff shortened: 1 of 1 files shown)


dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
```diff
@@ -1117,7 +1119,22 @@ public void startL1Listener(Set<ClientID> existingConnections) throws IOExceptio
...
+        this.l1Listener.start(existingConnections);
+        clientBound = true;
+      } catch (BindException bind) {
+  // this seems to happen on windows but should not.  we just gave up the port.  
+  // loop forever as a hack.
+        logger.warn("client server port not available:", bind);
+        try {
+          TimeUnit.SECONDS.sleep(1);
+        } catch (InterruptedException ie) {
+          throw new IOException(ie);
```
jd0-sag: Would it be more descriptive to re-throw the BindException, in this case?

jd0-sag:
This seems like a concise tactical solution so long as we can verify that it addresses the issue.

myronkscott:
updated considering comments

PR #512, opened 2017-03-17
EntityPersistor: synchronize all public methods

ramsai1729:


myronkscott:
EntityPersistor should only be accessed by the ProcessTransactionHandler or the ReplicatedTransactionHandler.  Single threaded.  Can you provide an example of multi-treaded access and maybe we can fix that.

myronkscott:
upon further inspection, it looks like storeToDisk method is accessed along the result path which is multi-threaded and I think unsafe regardless of CHM or HM.  I think I would rather have this class properly synchronized rather than use CHM.

myronkscott:
Hi Sairam,  

Can you please change this to use method synchronization in EntityPersistor?  This class is rarely accessed so synchronization is more straightforward in this case.

PR #518, opened 2017-03-24
Add isShutdown

henri-tremblay:
It is useful to know if a connection was closed. So we should expose this attribute.

### DIFF
--- a/connection-impl/src/main/java/com/terracotta/connection/TerracottaConnection.java
+++ b/connection-impl/src/main/java/com/terracotta/connection/TerracottaConnection.java
@@ -47,6 +47,10 @@ public class TerracottaConnection implements Connection {
     this.shutdown = shutdown;
   }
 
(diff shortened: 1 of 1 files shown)


jd0-sag:
This tightly couples the API to a detail of the implementation.

I question how you want to use this information, as well.  Where do you have access to a `TerracottaConnection` instance without a downcast?

henri-tremblay:
I was needing it when implementing a `ClusterManager`. To know if the cluster is active. So yes I do downcast.

myronkscott:
Not sure of the exact usage here but there are several implementations of ```Connection``` so ClusterManager may not always be able to downcast.  I don't think we should do this.  It would be better for the caller of ```Connection.close()``` to track state.  Closing this PR.  If you absolutely require Connection to expose state, please file an issue in terracotta-apis as an API discussion

Issue #521, opened 2017-03-30
Add missing methods to Diagnostics


anthonydahanne:
It would be ideal to have those missing methods added to the [Diagnostics](https://github.com/Terracotta-OSS/terracotta-core/blob/master/connection-loader/src/main/java/com/terracotta/diagnostic/Diagnostics.java) entity , that is already available from the diagnostic:// scheme : 

- String takeThreadDump()
- String getVMProperties()
- void terminateServer()

PR #528, opened 2017-04-05
DO NOT MERGE Issue #521 : Add methods to Diagnostic entity

anthonydahanne:
This PR is for initial discussion purposes.

Method calls from the client seem to work fine (tested locally)

My goal was to explore what was accessible, and how - no to provide a final implementation.
Comments are welcome, especially if you see something you think should not be accessible through the Diagnostic entity (for technical reasons for example)

### DIFF
--- a/connection-loader/src/main/java/com/terracotta/diagnostic/Diagnostics.java
+++ b/connection-loader/src/main/java/com/terracotta/diagnostic/Diagnostics.java
@@ -23,6 +23,21 @@ import org.terracotta.connection.entity.Entity;
 
 public interface Diagnostics extends Entity {
   String getState();
-  
-  String getClusterState();  
+
+  String getClusterState();
+
+  String getConfig();
+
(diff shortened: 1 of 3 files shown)


myronkscott:
I think there is going to be support for JMX in diagnostics, it should be supported more broadly so that adding items added in the future will not have to touch the handler.  I think there may be security concerns with such broad support however to how this is implemented needs to be carefully considered.

jd0-sag:
The part of this which seems particularly concerning is the logging hook.  How much data will that keep in memory?  Is that too much to remain stable (wasting/overflowing heap space) or too little to be meaningful (what question is it trying to answer, anyway)?

I guess I want to know what actual data is expected when the request for logs is issued.  What specific problem is being solved and can it be answered in any way other than them fetching the log files from the machine, directly?

myronkscott:
merged alternate PR

PR #533, opened 2017-04-12
Add more diagnostic methods

myronkscott:
add the following raw methods to ```Diagnostics```
```
  String getConfig();
  String getProcessArguments();
  String getServerInfoAttributes();
  String takeThreadDump();
  String terminateServer();
  String forceTerminateServer();  
```  

add support to call MBean methods with no args and get and set attributes


myronkscott:
NOT READY FOR MERGE YET

anthonydahanne:
hello @myronkscott , that looks good (I haven't run it yet, just read it)
I have 2 remarks though : 

1. I forgot to tell you about 2 methods that we would potentially want : 
```getTCProperties()```
and
```getEnvironment()```
Now those 2 are available from the TCServerInfoMBean, and I have noticed you have provided open access to 3 MBeans, including TCServerInfoMBean; so I guess you have the choice of implementing them in the handler somehow OR of telling me to use the Mbean access.

2. @mathieucarbou and I discussed a bit about how we're gonna handle Diagnostic connections. With my current implementation, and I believe with yours as well, calling ```forceTerminateServer``` from a fetched entity : 
```
     Connection connection = diagnosticConnectionService.connect(serverDiagnosticUri, properties);
     EntityRef<Diagnostics, Object, Void> ref = connection.getEntityRef(Diagnostics.class, 1, "root");
     Diagnostics d = ref.fetchEntity(null);
     d.forceTerminateServer()
```

is hanging the client : the call to d.forceTerminateServer() would never return, while the TC Server is properly killed.
We could wrap the call in a timing out thread somehow, but we don't believe we'd be able to kill the connection properly anyway.
So, we think that you could wrap all method returns in Future<> , such as : 
```
  Future<String> getState();
  Future<String> getClusterState();
etc.
```
so that we could properly handle things like 
```d.forceTerminateServer().get(10, TimeUnit.SECONDS)```

Tell me what you think, and thanks again for addressing this issue !

**Comment by myronkscott** [comment]
For 1, use the MBean access.  ```d.get(Server,TCProperties)```

For 2, the client no longer waits for a reply when terminating so it should not block.

This should now be good to merge.

**Comment by anthonydahanne** [comment]
thanks, testing it out right now - in particular the shutdown

**Comment by myronkscott** [comment]
changed ```takeThreadDump()``` to ```getThreadDump()```

**Comment by anthonydahanne** [comment]
I just had a look, and tried out forceTerminateServer, the client did not hang.
good to go for me

**Comment by jd0-sag** [comment]
My only concerns with any of this are these:

- is it sufficient to use space-delimited arguments or do we need some higher-level marshalling mechanism?
- is it safe to expose the ability to call methods on the MBean or do we need to restrict this in any way?

**Comment by anthonydahanne** [comment]
oh wait @myronkscott did you remove getServerInfoAttributes ?
that was convenient to regroup 4 common info under the same command...

**Comment by myronkscott** [comment]
I'm concerned about message size and formatting.  Can you do 4 get calls on the client side and basically do the aggregation there?

**Comment by anthonydahanne** [comment]
I can. That just does not sound efficient connection-wise. 
Anyway, we'll see how it goes.

**Comment by myronkscott** [comment]
Are you opening a separate connection for every call?  It should just be 4 successive calls to the Diagnostic entity.  It may be better because message size is smaller.

File: connection-loader/src/main/java/com/terracotta/diagnostic/DiagnosticEntityClientService.java [hunk 1]
```diff
@@ -52,15 +53,27 @@ public Diagnostics create(final EntityClientEndpoint<EntityMessage, EntityRespon
...
+              } else {
+                return methodName;
+              }
             }
-
-          }).invoke().get().toString();
+          }).invoke();
+          // if the server is terminating, never going to get a message back.  just return null
+          if (!methodName.equals("terminateServer") && !methodName.equals("forceTerminateServer")) {
+            return returnValue.get().toString();
```
mathieucarbou: @myronkscott @anthonydahanne : did you talk about the fact that `Diagnostics` method calls should not block and we should be able as a client to use them with timeout, to avoid blocking the health service and user requests ?
We could wrap the InvokeFuture to a Future like it's done in the voltrn proxy and have the `Diagnostics` interface return `Future<Foo>` instead.

anthonydahanne:
ah yes you're right. Haven't implemented client side yet. 
But yeah, 4 calls in a raw to the same fetched entity

connection-loader/src/main/java/com/terracotta/diagnostic/DiagnosticEntityClientService.java
myronkscott: I do not understand the request.  Diagnostics is supposed to be very simple request/response.  It has nothing to do with user requests or health services.  Remember, diagnostics connection is its own thing.  You cannot fetch the DiagnosticEntity from a regular connection.

anthonydahanne: yes @mathieucarbou ; @myronkscott added this code to handle the shutdown on the client (not hanging it)

mathieucarbou: @myronkscott : in a traditional connection, if for example we were to do a call to `getConfig()` and the active crashed, then the call will block.

If we connect with diagnostic and we do a call to a method of this entity, but the server crash or something wrong happens during this time.

What will be the behavior ? The call will block, until it can be fulfilled ?

myronkscott:
I added a timeout on requests that can be configured.

connection-loader/src/main/java/com/terracotta/diagnostic/DiagnosticEntityClientService.java
```diff
@@ -49,7 +52,9 @@ public Object deserializeConfiguration(byte[] bytes) {
   }
 
   @Override
-  public Diagnostics create(final EntityClientEndpoint<EntityMessage, EntityResponse> ece, Void ignore) {
+  public Diagnostics create(final EntityClientEndpoint<EntityMessage, EntityResponse> ece, Properties possible) {
```
anthonydahanne: possible ? that's a strange name for the properties

myronkscott: someone may not include it.  it is self contained and internal

myronkscott:
To address Jeff's concerns, I restricted the MBean access to only the terracotta domain.  Care must be taken when implementing new MBean features in this domain that it must not contain whitespace if it is to be accessed via Diagnostic

connection-loader/src/main/java/com/terracotta/diagnostic/DiagnosticEntityClientService.java
anthonydahanne: ok

anthonydahanne:
@myronkscott @jd0-sag ready to merge ?

anthonydahanne:
thanks

PR #559, opened 2017-05-12
Fix #509: propagate EntityUserException to client

### LINKED ISSUE #509: Investigate Entity Exception propagation to the client
If an entity server component throws an exception, it will bring down the server.  Investigate a system of exception propagation to the client in a standard way.

ramsai1729:
*  EntityUserException(s) thrown by ActiveServerEntity.invoke() will be logged and sent to the client and PassiveServerEntity.invoke() will be just logged. 
   * EntityUserException(s) don't crash the server but other exceptions crash and they will be not sent to the client.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,7 +41,7 @@
     <tc-shader.version>1.2</tc-shader.version>
     <skip.deploy>false</skip.deploy>
 
-    <terracotta-apis.version>1.3.0-pre3</terracotta-apis.version>
+    <terracotta-apis.version>1.3.0-pre4</terracotta-apis.version>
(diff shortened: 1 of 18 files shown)


pom.xml
```diff
@@ -41,7 +41,7 @@
     <tc-shader.version>1.2</tc-shader.version>
     <skip.deploy>false</skip.deploy>
 
-    <terracotta-apis.version>1.3.0-pre3</terracotta-apis.version>
+    <terracotta-apis.version>1.3-SNAPSHOT</terracotta-apis.version>
```
myronkscott: no snapshot releases.  adding a reminder

ramsai1729:
will push rebased commit after getService() related updates pushed to this repo.

PR #565, opened 2017-05-17
Fix #564 - fix group batch exception handling

### LINKED ISSUE #564: Group batching can get stuck with exception
exceptions sending in batch to other servers can get stuck if a network exception occurs because the exception is never cleared.

myronkscott:


myronkscott:
PR build failing due to flakey test a fix point.  Ignore cloudbees

PR #576, opened 2017-05-24
Fix #575 - Update to more strict dependencies for galvan support

### LINKED ISSUE #575: Clean up dependencies in galvan-support
Currently `galvan-support` pulls `terracotta-kit` as a `compile` dependency when it is needed only for `test`.
It uses `client-runtime` but this should be moved to `provided` as the exact client library is a factor of where the test is run.
Also the `junit` dependency can be moved to `provided` as it is clear usage of `galvan-support` requires `junit`

myronkscott:


galvan-support/pom.xml
```diff
@@ -46,12 +47,14 @@ limitations under the License.
...
       <artifactId>client-runtime</artifactId>
       <version>${project.version}</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.terracotta.internal</groupId>
       <artifactId>terracotta-kit</artifactId>
       <type>zip</type>
       <version>${project.version}</version>
+      <scope>test</scope>
```
ljacomet: Created #575 where I realised this dependency can be fully removed so I suggest either to update this PR or to merge the other one.

myronkscott:
Closed in favor of #577

PR #579, opened 2017-05-30
Fix #578 Update reference directory for server

### LINKED ISSUE #578: Enable moving the server from the kit to different locations
Currently the way the `start-tc-server` scripts are written, if you move the `server` folder from the kit to a different location, it will only work if you keep the `server` name.

We should make it possible to move around the `server` folder and renaming it as a way to copy the server binaries on multiple machines without having to keep the full kit structure.

ljacomet:
Move it to the server folder instead of the kit root. This allows
moving out server from a kit and use that as the set of files to copy
on different hosts.

### DIFF
--- a/common/src/main/java/com/tc/config/Directories.java
+++ b/common/src/main/java/com/tc/config/Directories.java
@@ -60,18 +60,15 @@ public class Directories {
   /**
    * Get installation root directory.
    * 
-   * @return Installation root directory or null if TC_INSTALL_ROOT_IGNORE_CHECKS_PROPERTY_NAME is set and
+   * @return Installation root directory or {@code user.dir} if TC_INSTALL_ROOT_IGNORE_CHECKS_PROPERTY_NAME is set and
    *         TC_INSTALL_ROOT_PROPERTY_NAME is not.
-   * @throws FileNotFoundException If {@link #TC_INSTALL_ROOT_PROPERTY_NAME} has not been set. If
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/tc/config/Directories.java
```diff
@@ -60,18 +60,15 @@
...
+   *         this exception may be thrown if the installation root directory is not a directory
    */
-  public static File getInstallationRoot() throws FileNotFoundException {
+  static File getInstallationRoot() throws FileNotFoundException {
     boolean ignoreCheck = Boolean.getBoolean(TC_INSTALL_ROOT_IGNORE_CHECKS_PROPERTY_NAME);
     if (ignoreCheck) {
-      // XXX hack to have enterprise system tests to find license key under <ee-branch>/code/base
-      String baseDir = System.getProperty("tc.base-dir");
-      return new File(baseDir != null ? baseDir : ".", "../../../code/base");
+      return new File(System.getProperty("user.dir"));
```
ramsai1729: I don't think we are using this property `TC_INSTALL_ROOT_IGNORE_CHECKS_PROPERTY_NAME`, its better to remove it altogether instead of patching it.

ljacomet: It is still configured in `build-parent/pom.xml` and I did not want to start a big clean up here, just a minor removal. Filed #582 for this bigger clean up.

PR #580, opened 2017-05-30
[No Merge] Prototype - Add client state to cluster state dump 

ramsai1729:
This PR prototypes adding client state to cluster state dump, dumping thread state looks okay but entity state and manager state is messy, I have created separate commits. Another option we can explore is using JMX for dumping client state. 

Please have a look.


myronkscott:
moved to   Terracotta-OSS/terracotta-platform#331

PR #581, opened 2017-05-30
Terracotta server PID file creation/deletion using startup scripts

saurabhagas:
This PR contains the .bat file changes only. sh file changes will be on similar lines, and can be done quickly. I can update the PR with sh changes as well once the overall philosophy and approach is approved.


terracotta-org:

GaryWKeim:
This appears (not a BAT expert) to only support a single running server at a time, no?
I'm not clear on the requirements... maybe that's fine.

PR #586, opened 2017-06-02
issue-585: Reduce INFO level logging of Terracotta client

albinsuresh:


myronkscott:
Isn't a lot of this code shared between client and server?  Reducing will impact both.  Are you sure that none of this is needed at the info level on the server?

PR #587, opened 2017-06-02
Fix #549: Use server default port when port is not specified in conne…

### LINKED ISSUE #549: Make port in connection URI optional
It seems that even though we have on the server side a concept of default port - `9510` for now - this is not something we can rely on when opening a connection.
The following connection URI causes and `AssertionError`: `terracotta://localhost` even when a server on `9510` is started on the machine.

Could support for making the port optional in the connection URI be added?

ramsai1729:
…ction URI

### DIFF
--- a/dso-l1/src/main/java/com/tc/config/schema/L2ConfigForL1.java
+++ b/dso-l1/src/main/java/com/tc/config/schema/L2ConfigForL1.java
@@ -37,7 +37,7 @@ public interface L2ConfigForL1 {
     public L2Data(String host, int tsaPort, boolean secure) {
       Assert.assertNotBlank(host);
       this.host = host;
-      this.tsaPort = tsaPort;
+      this.tsaPort = tsaPort < 0 ? 9410 : tsaPort; // use default port if given one is not valid
       this.secure = secure;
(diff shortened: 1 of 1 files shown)


dso-l1/src/main/java/com/tc/config/schema/L2ConfigForL1.java
```diff
@@ -37,7 +37,7 @@ public L2Data(String host, int tsaPort) {
     public L2Data(String host, int tsaPort, boolean secure) {
       Assert.assertNotBlank(host);
       this.host = host;
-      this.tsaPort = tsaPort;
+      this.tsaPort = tsaPort < 0 ? 9510 : tsaPort; // use default port if given one is not valid
       this.secure = secure;
```
myronkscott: Shouldn't you use the new default port here?

ramsai1729: its the correct port now, can be changed after new tc-config release.

myronkscott:
please rebase

ramsai1729:
rebased

PR #588, opened 2017-06-02
Adapt tc config default port changes

ramsai1729:
Should follow https://github.com/Terracotta-OSS/terracotta-configuration/pull/58 merge 

Note that 

- there are still some tests which references 9510 port (or other default port numbers) but they are explicitly configured in the test xml or code. 
- ConnectionInfoConfig still uses hardcoded tsa port as dso-l1 doesn't depend on terracotta-configuration, would like discuss internally before adding that dependency
- ConnectionInfoConfig's code that references hardcoded tsa port looks like a dead code but we need this dependency change for Terracotta-OSS/terracotta-core#549


myronkscott:
terracotta-configuration 10.3.0-pre6 released.  Please update and rebase.

PR #592, opened 2017-06-08
Disable health checking by default

ChrisGreenaway:


GaryWKeim:
Can you say why this is or refer to the BF?

ChrisGreenaway:
Not now.

PR #603, opened 2017-06-22
Replace TCLogging with slf4j + logback

albinsuresh:


dso-l2/src/main/java/com/tc/l2/logging/TCLogbackLogging.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  public static void redirectLogging(String logDir) {
+    LoggerContext loggerContext = (LoggerContext) LoggerFactory.getILoggerFactory();
+    ch.qos.logback.classic.Logger internalLogger = loggerContext.getLogger(Logger.ROOT_LOGGER_NAME);
+    Appender<ILoggingEvent> appender = internalLogger.getAppender("TC_BASE");
+    if (appender != null) {
+      String logLocation = logDir + "/terracotta.server.log";
+      LOGGER.info("Log file: {}", logLocation);
+      BufferingAppender<ILoggingEvent> base = (BufferingAppender) appender;
```
ChrisGreenaway: What happens if the cast fails?

tc-messaging/src/main/java/com/tc/logging/TCLoggingService.java
```diff
@@ -1,32 +0,0 @@
...
- *
- *  The Initial Developer of the Covered Software is
- *  Terracotta, Inc., a Software AG company
- *
- */
-package com.tc.logging;
-
-import java.net.URI;
-
-public interface TCLoggingService {
```
ChrisGreenaway: Are we not having all these different categories of logger any more?

dso-l1/src/main/java/com/tc/config/schema/setup/ClientConfigurationSetupManagerFactory.java
```diff
@@ -34,7 +34,7 @@
  */
 public class ClientConfigurationSetupManagerFactory {
 
-  private static final TCLogger consoleLogger = CustomerLogging.getConsoleLogger();
```
albinsuresh: There won't be any console logging on the client side anymore as it is up to the user to configure what goes to console and what not. We won't be enforcing anything. That's why specific console loggers were removed from all client side classes.

dso-l1/src/main/java/com/tc/object/DistributedObjectClient.java
```diff
@@ -133,10 +134,7 @@
  */
 public class DistributedObjectClient implements TCClient {
 
-  protected static final TCLogger                    DSO_LOGGER                          = CustomerLogging
-                                                                                             .getDSOGenericLogger();
-  private static final TCLogger                      CONSOLE_LOGGER                      = CustomerLogging
-                                                                                             .getConsoleLogger();
+  protected static final Logger DSO_LOGGER = LoggerFactory.getLogger(DistributedObjectClient.class);
```
albinsuresh: Again another client side class that had logging to both console and the user's target. No more console logging here either.

dso-l1/src/main/java/com/tc/object/config/ConnectionInfoConfig.java
```diff
@@ -32,7 +33,7 @@
  * Returns a {@link ConnectionInfo} array from the L2 data.
  */
 public class ConnectionInfoConfig {
-  static TCLogger consoleLogger = CustomerLogging.getConsoleLogger();
```
albinsuresh: Again: No more console logging for client.

dso-l2/src/main/java/com/tc/l2/logging/BufferingAppender.java
```diff
@@ -18,43 +18,32 @@
...
+    this.buffer = new ConcurrentLinkedQueue<>();
+    this.on = true;
   }
 
   @Override
-  public void close() {
-    // nothing needs to be here.
+  public void doAppend(E eventObject) {
+    buffer.add(eventObject);
+    super.doAppend(eventObject);
```
albinsuresh: This appender is an extension of ConsoleAppender. So while the buffering is happening on bootstrap we send those to the console as well.

dso-l2/src/main/java/com/tc/l2/logging/TCLogbackLogging.java
albinsuresh: Currently I'm relying on a special appender named "TC_BASE" which is of type `BufferingAppender`. But nothing is stopping a user from overriding with his config having the same name. So yeah I guess it needs to be handled better to be on the safer side.

tc-messaging/src/main/java/com/tc/logging/TCLoggingService.java
albinsuresh: I didn't quite get that comment. Which category are you talking about?

dso-l2/src/main/java/com/tc/l2/logging/StartupAndSizeBasedTriggeringPolicy.java
```diff
@@ -0,0 +1,17 @@
...
+import ch.qos.logback.core.rolling.SizeBasedTriggeringPolicy;
+
+import java.io.File;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class StartupAndSizeBasedTriggeringPolicy<E> extends SizeBasedTriggeringPolicy<E> {
+    private final AtomicBoolean firstTime = new AtomicBoolean();
+
+    public boolean isTriggeringEvent(final File activeFile, final E event) {
+        if (firstTime.compareAndSet(false, true) && activeFile != null && activeFile.length() > 0) {
```
albinsuresh: Actually there isn't a case that I can see when the `activeFile` is null. So I guess those checks can be removed.

common/src/main/java/com/tc/async/impl/StageManagerImpl.java
```diff
@@ -89,7 +90,7 @@ private void printStats() {
             stat.logDetails(logger);
           }
         } catch (Throwable th) {
-          logger.error(th);
+          logger.error("Exception :", th);
```
albinsuresh: The whole reason why this prefix was added was because of slf4j's lack of support for a logger api that just takes a `Throwable`. They force us to provide a message. But we never had one with TCLogging. We were relying on the exception messages for details and I just left it like that.

common/src/main/java/com/tc/logging/TCLoggerProvider.java
```diff
@@ -18,7 +18,9 @@
  */
 package com.tc.logging;
 
+import org.slf4j.Logger;
+
 public interface TCLoggerProvider {
```
albinsuresh: No this is used. This class has nothing to do with old TCLogging. It's just in the name.

PR #607, opened 2017-06-26
Transport close

myronkscott:
transport close is the end of any client.  Make sure that event is propagated to a client shutdown.


cschanck:
Looks right to me -- why the failure?

myronkscott:
timing test on a slow machine.  I might as well try and better the test

PR #612, opened 2017-06-28
Refactor persistors to be more flexible and optimized based on client…

myronkscott:
… type


dso-l2/src/test/java/com/tc/objectserver/persistence/PersistorTest.java
```diff
@@ -0,0 +1,117 @@
...
+  public void setUp() {
+  }
+  
+  @After
+  public void tearDown() {
+  }
+
+  @Test
+  public void testStart() throws Throwable {
+    IPlatformPersistence persist = mock(IPlatformPersistence.class);
```
cschanck: Mock, mock, mock. :-)

PR #617, opened 2017-07-06
cleanup info logs

ramsai1729:
This PR removes some unnecessary logging at info level and rewording some logs.

### DIFF
--- a/common/src/main/java/com/tc/net/protocol/transport/ConnectionPolicyImpl.java
+++ b/common/src/main/java/com/tc/net/protocol/transport/ConnectionPolicyImpl.java
@@ -18,8 +18,6 @@
  */
 package com.tc.net.protocol.transport;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import com.tc.util.Assert;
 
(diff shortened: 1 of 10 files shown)


common/src/main/java/com/tc/net/protocol/transport/ServerStackProvider.java
```diff
@@ -393,7 +393,7 @@ private void handleSyn(SynMessage syn) throws RejectReconnectionException, Produ
           }
         }
       }
-      logger.info("User " + principal + " successfully authenticated");
+      //logger.info("User " + principal + " successfully authenticated");
```
myronkscott: remove

common/src/main/java/com/tc/net/protocol/transport/ConnectionPolicyImpl.java
```diff
@@ -72,19 +72,19 @@ public synchronized boolean connectClient(ConnectionID connID) {
     HashSet<ConnectionID> jvmClients = clientsByJvm.get(connID.getJvmID());
 
     if (isMaxConnectionsReached() && jvmClients == null) {
-      logger.info("Rejecting " + connID + "; " + toString());
+      //logger.info("Rejecting " + connID + "; " + toString());
       return false;
```
myronkscott: remove all the commented out log statements in this file

ramsai1729: yeah, can be removed. kept it there just in case this code gets reused.

common/src/main/java/com/tc/net/protocol/transport/ServerStackProvider.java
ramsai1729: okay

PR #624, opened 2017-07-12
Fix client persistor

myronkscott:


dso-l2/src/main/java/com/tc/objectserver/handler/RetirementManager.java
```diff
@@ -103,6 +104,7 @@ public synchronized void registerWithMessage(EntityMessage invokeMessage, int co
     List<Retiree> toRetire = new ArrayList<>();
 
     LogicalSequence completedRequest = this.currentlyRunning.remove(completedMessage);
+    this.mostRecentRegisteredToKey.remove(completedRequest.concurrencyKey, completedRequest);
```
ramsai1729: Why do we need to do this?

dso-l2/src/test/java/com/tc/objectserver/handler/RetirementManagerTest.java
```diff
@@ -317,6 +317,36 @@ public void testRetirementWithMultiMessageDependency() throws Exception {
...
+    
+    this.retirementManager.deferRetirement(invokeMessage2, invokeMessage3);
+    
+    registerWithMessage(invokeRequest2, invokeMessage2, 1);
+
+    registerWithMessage(invokeRequest3, invokeMessage3, 1);
+    
+    List<Retiree> retireList = this.retirementManager.retireForCompletion(invokeMessage1);
+    
+    retireList = this.retirementManager.retireForCompletion(invokeMessage2);
```
ramsai1729: Assert on this list would be helpful.

ramsai1729: Also it would be to good if you can add some test with messages on different concurrency keys.

dso-l2/src/main/java/com/tc/objectserver/handler/RetirementManager.java
myronkscott: Minor cleanup.  I don't want things hanging around in the map.  For example, if an entity decided to use a random int for the concurrency key, this would prevent a leak

dso-l2/src/test/java/com/tc/objectserver/handler/RetirementManagerTest.java
myronkscott: good call.   will fix

myronkscott:
Updated and rebased

ramsai1729:
test this please

PR #633, opened 2017-07-14
Do not append sub directory to user provided log path

albinsuresh:


ljacomet:
Failed test is a known unstable one ...

PR #637, opened 2017-07-17
Fix #613: Add extended configurations state to cluster state dump

### LINKED ISSUE #613: Support implementing StateDumpable on server config plugins
Server plugins (offheap, dataroot) can hold some states that are useful for management and platform would need to also dump all plugins that implements `StateDumpable`

For the moment, as a workaround, I added some code in the monitoring service that will dump all `StateDumpable` plugins.

See: https://github.com/Terracotta-OSS/terracotta-platform/pull/343

ramsai1729:
updated the PR with new terracotta-configuration 10.3.0-pre13 release


myronkscott:
terracotta-configuration 10.3.0-pre13

PR #647, opened 2017-07-20
Fix #645: Propagate actual cause with TCShutdownServerException excep…

### LINKED ISSUE #645: Exception swallowed if entity creation fails during passive sync
The code in `ManagedEntityImpl#receiveSyncCreateEntity` creates an exception that does not carry over the root cause.

And the callback on exit in `DistributedObjectServer#start` only logs the exception message.

The root cause should be passed in the `TCShutdownServerException` and the log should contain the full exception stack trace.

ramsai1729:
…tion

### DIFF
--- a/common/src/main/java/com/tc/exception/TCShutdownServerException.java
+++ b/common/src/main/java/com/tc/exception/TCShutdownServerException.java
@@ -26,4 +26,8 @@ public class TCShutdownServerException extends TCRuntimeException {
   public TCShutdownServerException(String message) {
     super(message);
   }
+
+  public TCShutdownServerException(String message, Throwable cause) {
+    super(message, cause);
+  }
 }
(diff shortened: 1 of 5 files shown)


tc-messaging/src/main/java/com/tc/object/EntityID.java
```diff
@@ -84,9 +84,6 @@ public static EntityID readFrom(TCByteBufferInput serialInput) throws IOExceptio
 
   @Override
   public String toString() {
-    return "EntityID{" +
-           "className='" + className + '\'' +
-           ", entityName='" + entityName + '\'' +
-           '}';
+    return className + ":" + entityName;
   }
```
myronkscott: I'm not sure I like changing this.

myronkscott: grab the classname and entity name if you want this format

ramsai1729: not sure I understand this, could you tell me what exactly needs to be changed here?

myronkscott: remove this change.  If you want this format in code somewhere just produce it from the component parts.

ramsai1729: done

PR #656, opened 2017-07-25
Fix #653: catch exceptions on diagnostic commands code path

### LINKED ISSUE #653: no calls on the diagnostic port should crash the server
Currently, entity plugin code is called when a cluster state dump is taken.  If there is a fatal error, in this entity code, it can crash the server.  Make sure this call and any other call on the diagnostic port does not crash the server under any circumstance.

ramsai1729:


dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
```diff
@@ -327,13 +327,17 @@ protected ServerBuilder getServerBuilder() {
...
+      this.persistor.prettyPrint(pp);
+      this.groupCommManager.prettyPrint(pp);
+      this.l2Coordinator.prettyPrint(pp);
+      this.entityManager.prettyPrint(pp);
+      this.serviceRegistry.prettyPrint(pp);
+      addExtendedConfigState(pp);
+    } catch (Throwable t) {
+      pp.println("unable to collect complete cluster state due to " + t.getLocalizedMessage());
+    }
     return pp.toString().getBytes(set);
```
myronkscott: I would consider wrapping each sub-section

ramsai1729: I thought that was not necessary as we are already catching exceptions from entity code run in ManagedEntityImpl and TerracottaServiceProviderRegistryImpl and platform code where we add cluster state for other components looks fine. Anyways I will push necessary changes.

PR #663, opened 2017-07-28
syntax and logical corrections

akomakom:
Made bash syntax consistent and corrected what appeared to be bugs - classpath quoting and generation.

### DIFF
--- a/terracotta-kit/src/assemble/server/bin/start-tc-server.sh
+++ b/terracotta-kit/src/assemble/server/bin/start-tc-server.sh
@@ -31,7 +31,7 @@ TC_SERVER_DIR=$(dirname "$(cd "$(dirname "$0")";pwd)")
 PLUGIN_LIB_DIR="${TC_SERVER_DIR}/plugins/lib"
 PLUGIN_API_DIR="${TC_SERVER_DIR}/plugins/api"
 
-if test \! -d "${JAVA_HOME}"; then
+if ! [ -d "${JAVA_HOME}" ]; then
   echo "$0: the JAVA_HOME environment variable is not defined correctly"
   exit 2
 fi
@@ -44,8 +44,8 @@ for JAVA_COMMAND_ARGS in \
     "-client -XX:MaxDirectMemorySize=9223372036854775807" \
     "-XX:MaxDirectMemorySize=9223372036854775807"
 do
-  "${JAVA_HOME}/bin/java" $JAVA_COMMAND_ARGS -version > /dev/null 2>&1
-  if test "$?" = "0" ; then break; fi
+    # accept the first one that works
+    "${JAVA_HOME}/bin/java" $JAVA_COMMAND_ARGS -version > /dev/null 2>&1 && break
 done
 
 function setPluginClasspath {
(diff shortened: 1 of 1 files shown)


terracotta-kit/src/assemble/server/bin/start-tc-server.sh
```diff
@@ -54,48 +54,46 @@ function setPluginClasspath {
...
 
- if test "$exitValue" = "11"; then
-   start=true;
-   echo "start-tc-server: Restarting the server..."
- else
-   exit $exitValue
- fi
+    if [ $exitValue -eq 11 ] ; then
+        echo "$0: Restarting server"
+        sleep 1
```
akomakom: I added sleep 1, simply because this has a high probability of busywaiting - but a configurable or higher delay or an iteration limit may be better.

terracotta-kit/src/assemble/server/bin/start-tc-server.sh
```diff
@@ -54,48 +54,46 @@ function setPluginClasspath {
...
-#  support services that may use SLF4j for logging
-            for jarFile in "${TC_SERVER_DIR}"/lib/slf4j*.jar
-            do
-                PLUGIN_CLASSPATH=${PLUGIN_CLASSPATH}:${jarFile}
+                PLUGIN_CLASSPATH="${PLUGIN_CLASSPATH}:${jarFile}"
             done
+
         fi
     done
+    #  Adding SLF4j libraries to the classpath of the server to
```
akomakom: I moved this out of the loop since it has no obvious relationship to either the loop or the conditional, and will cause the slf4j jars to be added twice

terracotta-kit/src/assemble/server/bin/start-tc-server.sh
```diff
@@ -54,48 +54,46 @@ function setPluginClasspath {
...
 
-PLUGIN_CLASSPATH=${PLUGIN_CLASSPATH}:${TC_SERVER_DIR}/lib
+PLUGIN_CLASSPATH="${PLUGIN_CLASSPATH}:${TC_SERVER_DIR}/lib"
 
 #rmi.dgc.server.gcInterval is set an year to avoid system gc in case authentication is enabled
 #users may change it accordingly
-start=true
-while "$start"
-do
+while [ 1 ] ; do
```
akomakom: the loop never exits naturally, so why bother with the condition...

terracotta-kit/src/assemble/server/bin/start-tc-server.sh
ramsai1729: do we even need to add slf4j api jar explicitly, looks like this gets added to classpath via tc.jar already 

````
Class-Path: dso-l2-5.3-SNAPSHOT.jar tc-config-parser-10.3.0-pre15.jar
 tcconfig-schema-10.3.0-pre15.jar logback-classic-1.1.3.jar logback-co
 re-1.1.3.jar slf4j-api-1.7.7.jar entity-server-api-1.3.0-pre12.jar en
 tity-common-api-1.3.0-pre12.jar standard-cluster-services-1.3.0-pre12
 .jar packaging-support-1.3.0-pre12.jar monitoring-support-1.3.0-pre12
 .jar common-5.3-SNAPSHOT.jar tc-messaging-5.3-SNAPSHOT.jar build-data
 -5.3-SNAPSHOT.jar management-5.3-SNAPSHOT.jar connection-api-1.3.0-pr
 e12.jar commons-cli-1.1.jar resources/
````

PR #668, opened 2017-07-31
Cleanup

myronkscott:


dso-l1/src/main/java/com/tc/object/EntityClientEndpointImpl.java
```diff
@@ -49,6 +53,7 @@
   private final Runnable closeHook;
   private EndpointDelegate delegate;
   private boolean isOpen;
+  private ExecutorService closer;
```
ramsai1729: this can be final

dso-l1/src/main/java/com/tc/object/EntityClientEndpointImpl.java
```diff
@@ -235,6 +241,28 @@ public void close() {
...
+    };
+    if (this.closer == null) {
+      close();
+      return new CompletedFuture();
+    } else {
+      try {
+        return this.closer.submit(call);
+      } catch (RejectedExecutionException re) {
+        // connection already shutdown
+        return new CompletedFuture(new IllegalStateException("connection has already been shutdown"));
```
ramsai1729: is this always true?

dso-l1/src/test/java/com/tc/object/ClientEntityManagerTest.java
```diff
@@ -243,6 +245,84 @@ public TCMessage answer(InvocationOnMock invocation) throws Throwable {
...
+    }
+    
+    // Now, release it and expect to see the exception thrown, directly (since we are accessing the manager, directly).
+    boolean didRelease = false;
+    try {
+      Future<Void> released = fetcher.release();
+      released.get();
+      didRelease = true;
+    } catch (Exception e) {
+      // Expected.
```
ramsai1729: looks like comments are not correct here, please fix

dso-l1/src/main/java/com/tc/object/ClientEntityManagerImpl.java
```diff
@@ -97,6 +99,7 @@
   private boolean isShutdown = false;
   private final boolean reconnectable;
 
+  private final ExecutorService endpointCloser = Executors.newCachedThreadPool();
```
ramsai1729: better to go with fixed number of threads here?

dso-l1/src/main/java/com/tc/config/schema/setup/L1ConfigurationSetupManager.java
```diff
@@ -39,4 +40,6 @@
   L2ConfigForL1 l2Config();
 
   SecurityInfo getSecurityInfo();
+  
+  Map<String, String> getOverrideTCProperties();
```
ramsai1729: is this change complete? don't see a way to pass tc properties

dso-l1/src/main/java/com/tc/object/EntityClientEndpointImpl.java
myronkscott: yes.  will fix

dso-l1/src/main/java/com/tc/object/EntityClientEndpointImpl.java
myronkscott: with the choice of the unbounded executor service, I believe this is always true.

dso-l1/src/main/java/com/tc/object/ClientEntityManagerImpl.java
myronkscott: i don't think so.  We do not want a delay is starting execution of this task.  I think however, that multiple calls to the same release should return the same future.  Will fix.

dso-l1/src/main/java/com/tc/config/schema/setup/L1ConfigurationSetupManager.java
myronkscott: Yes.  This is an old feature from when the clients had the server's tc-config.  I left it so that in the future, if we wanted a new way to add properties on the client, the hook would still be there.

PR #669, opened 2017-08-01
Configuration extension for server logging

albinsuresh:
This `logback-ext.xml` is intended to be used by developers. This file is packaged as part of the kit and is documented in itself. If we don't want this file exposed like this we could make it a hidden file and document the existence of such a file elsewhere. Otherwise we could completely avoid packaging this file and have this template also included in the documentation. Not sure what's the best thing to do here. So suggestions welcome.

### DIFF
--- a/dso-l2/src/main/resources/logback.xml
+++ b/dso-l2/src/main/resources/logback.xml
@@ -27,5 +27,7 @@
     <appender-ref ref="TC_BASE" />
   </logger>
 
+  <include optional="true" resource="logback-ext.xml"/>
+
 </configuration>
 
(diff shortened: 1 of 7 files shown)


myronkscott:
I'm surprised that this gets picked up properly there.  It would be nice to have an external location (user.home) (user.dir) so that galvan tests can be run with custom settings when needed.

albinsuresh:
That extension file gets picked up because server/lib path was added to class-path in the startup scripts when I did the original logging changes.

The second commit addresses server log customizability for galvan. Rather than having a generic path like user.home or user.dir that would affect all galvan tests run with that system I wanted to provide a more granular cluster specific configurability.

With this change a user can have an log extension file in the classpath with any random name of his choice and use that name in configuring a specific cluster. If someone wants a global config file that affects all tests all he has to do is add an extension file named "`logback-ext.xml` in the classpath and it'll be picked up by all tests for which an explicit log configuration is not there.

This PR requires https://github.com/Terracotta-OSS/galvan/pull/171 to be merged.

myronkscott:
update with galvan v1.3.0-pre9

myronkscott:
rebase please

myronkscott:
Please rebase

albinsuresh:
Rebased

PR #673, opened 2017-08-03
Fix #670 - Remove incorrect warnings about valid tc properties

### LINKED ISSUE #670: "server.entity.processor.thread" is flagged as an "undefined" tc property
When a value for `server.entity.processor.threads`  (`com.tc.properties.TCPropertiesConsts#ENTITY_PROCESSOR_THREADS`) is included in a `tc.properties` file, the following messages are logged:

```
2017-08-02 10:35:32,313 [main] WARN com.tc.properties.TCProperties - The property "server.entity.processor.threads" is not present in set of defined tc properties. Probably this is misspelled
2017-08-02 10:35:32,313 [main] INFO com.tc.properties.TCProperties - The property "server.entity.processor.threads" was overridden to 4 from null by the tc-config file
```

This property value is _successfully_ used in `com.tc.net.utils.L2Utils#getOptimalApplyStageWorkerThreads` so it should not be flagged as not defined.

ramsai1729:


common/src/main/java/com/tc/properties/TCPropertiesConsts.java
```diff
@@ -407,4 +387,15 @@
   public static final String  LOGGING_CATEGORY                                              = "logging";
   public static final String  NETCORE_CATEGORY                                              = "net.core";
 
+  String[] TC_PROPERTIES_WITH_NO_DEFAULTS = {
+      ENTITY_PROCESSOR_THREADS,
+      L2_TCCOM_WORKERTHREADS,
+      L2_SEDA_STAGE_WORKERTHREADS,
+      L2_NHA_TCGROUPCOMM_RECONNECT_L2PROXY_TO_PORT,
+      CLIENT_MAX_PENDING_REQUESTS,
```
ramsai1729: looks like some of these properties can be moved to tc.properties, can be fixed later though.

PR #677, opened 2017-08-08
Fix #672: Start JMX remote monitoring from the Diagnostic Port

### LINKED ISSUE #672: Start JMX remote monitoring from the Diagnostic Port
I would add the command to the DSO MBean

Hint:

https://stackoverflow.com/questions/27508185/is-it-possible-to-enable-remote-jmx-monitoring-programmatically

ramsai1729:


dso-l2/src/main/java/com/tc/stats/DSO.java
```diff
@@ -66,6 +70,7 @@
 public class DSO extends AbstractNotifyingMBean implements DSOMBean {
 
   private final static Logger logger = LoggerFactory.getLogger(DSO.class);
+  private final static int DEFAULT_JMX_REMOTE_PORT = 5050;
```
myronkscott: Probably need this to be configurable somehow

ramsai1729: yeah, is there a way to pass this port info from Diagnostics entity?

myronkscott: I'd use the setter/getter bean pattern on the MBean and use Diagnostics.set

myronkscott:
Can you add a test?  A unit test would suffice if JUnit will let you do it.

dso-l2/src/main/java/com/tc/stats/DSO.java
myronkscott: you could even just start the server from the set method

PR #679, opened 2017-08-08
Improve message passing perf

myronkscott:


common/src/main/java/com/tc/net/core/CoreNIOServices.java
```diff
@@ -402,13 +402,14 @@ void addSelectorTask(Runnable task) {
...
         while (true) {
-          try {
-            this.selectorTasks.put(task);
-            break;
-          } catch (InterruptedException e) {
-            logger.warn("Exception: ", e);
-            isInterrupted = true;
+          while (!this.selectorTasks.offer(task)) {
+            try {
+              Thread.sleep(1);
```
cschanck: Really, 1 second?

PR #681, opened 2017-08-10
Filter logging for internal clients

myronkscott:


common/src/main/java/com/tc/net/protocol/transport/MessageTransportBase.java
```diff
@@ -405,5 +403,14 @@ public synchronized void setRemoteCallbackPort(int remoteCallbackPort) {
...
   }
+  
+  void log(String msg) {
+    if (!getConnectionId().getProductId().isInternal()) {
+      getLogger().info(msg);
+    } else {
+      getLogger().debug(msg);
+    }
+
+  }
```
cschanck: Could we factor out ProductId as an argument, make this a static and use it everywhere? Bunches of other files have the same pattern. Not sure if it is worth it, but...

PR #686, opened 2017-08-22
Stage shutdown

myronkscott:
Make the ClientEntityManager shutdown more friendly to consumers
Remove thread interrupts from Stage destroy
cleanup comm stack logging

currently adding tests so do not merge

### DIFF
--- a/common/src/main/java/com/tc/async/impl/AbstractStageQueueImpl.java
+++ b/common/src/main/java/com/tc/async/impl/AbstractStageQueueImpl.java
@@ -4,24 +4,51 @@ import org.slf4j.Logger;
 
 import com.tc.async.api.EventHandler;
 import com.tc.async.api.EventHandlerException;
+import com.tc.async.api.Sink;
 import com.tc.async.api.Source;
 import com.tc.async.api.SpecializedEventContext;
 import com.tc.async.api.StageQueueStats;
-import com.tc.exception.TCRuntimeException;
-import com.tc.util.UpdatableFixedHeap;
-
-import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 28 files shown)


common/src/main/java/com/tc/async/impl/AbstractStageQueueImpl.java
```diff
@@ -4,24 +4,51 @@
...
+  }
+  
+  boolean isClosed() {
+    return closed;
   }
 
+  @Override
+  public void setClosed(boolean closed) {
+    this.closed = closed;
+    if (closed) {
```
cschanck: what if it is already closed?

myronkscott: It can't happen because StageImpl protects it.  It's  seems like a bad interdependency between the stage and the queue but i'm not sure how to fix it.  Maybe assert there are no double closes or opens

cschanck: I was just thinking to skip the setClosed path if it is already closed? So extra closes can't hurt it?

myronkscott: I think I'm going to change the API to close().  That is the way it is used and tested

cschanck: k

myronkscott: Stages and their queues are create -> destroy and not reusable the way the API and tests are laid out so I changed the setClosed method to just close() and once and only once assertion.

Issue #694, opened 2017-08-30
Add a meaningful log message when diagnostics handler fails


rkavanap:
When generic services are written that are not marked with BuiltIn, it may deploy those services only if the services are configured within platform. And if these services are not deployed, the corresponding JMX attributes and methods of that service may not be registered. 

This issue is for tracking of printing a meaningful warning message (in addition to what is already logged) to the log file, so that the end user (such as an admin) will get a clue about the failure reason.

myronkscott:
PR #697

PR #695, opened 2017-08-30
i-694: Log warning messages on server if JMX subsystem is not found

rkavanap:
Log a warning message if JMX subsystem throws exception.

This will help identify issues on the server if non-built in services are configured. This will also help in troubleshooting as well as unintended invocations from external tools.


rkavanap:
@myronkscott, Please merge if found Ok. I have locally tested and found the error messages to be satisfactory.

ljacomet:
This should be daggy fixed against `release/5.3` and master. Also please reference the issue in the commit comment using the `#694` notation

PR #700, opened 2017-09-05
ReplicatedTransactionHandler: call endSyncConcurrencyKey before invok…

ramsai1729:
…ing deferred messages


ramsai1729:
this will be fixed in PR https://github.com/Terracotta-OSS/terracotta-core/pull/696

PR #722, opened 2017-09-22
[Not for merge] Changes for fixing SyncState multi-threaded access issues

ramsai1729:
This work is not completed, created this PR to discuss the approach.

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/entity/ActiveToPassiveReplication.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/entity/ActiveToPassiveReplication.java
@@ -21,6 +21,7 @@ package com.tc.objectserver.entity;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.tc.async.api.Sink;
 import com.tc.l2.ha.L2HAZapNodeRequestProcessor;
 import com.tc.l2.msg.ReplicationAckTuple;
 import com.tc.l2.msg.ReplicationMessageAck;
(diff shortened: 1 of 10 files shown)


ramsai1729:
test this please

dso-l2/src/main/java/com/tc/objectserver/entity/ActiveToPassiveReplication.java
```diff
@@ -239,13 +247,19 @@ public ActivePassiveAckWaiter replicateActivity(SyncReplicationActivity activity
...
+              boolean isNormalComplete = true;
+              internalAckCompleted(activityID, node, null, isNormalComplete);
+            }
+            barrierCompletion.complete();
+          };
           // This isn't local-only so try to replicate.
-          didSend = this.replicationSender.replicateMessage(node, activity);
-        }
-        if (!didSend) {
+          this.replicationSenderMessageSink.addSingleThreaded(new SyncActivityMessage(node, activity, consumer));
```
myronkscott: We don't want to do this here.  This will slow replication way down.  What is the purpose of waiting for completion?

dso-l2/src/main/java/com/tc/objectserver/entity/ActiveToPassiveReplication.java
```diff
@@ -263,7 +277,9 @@ public void removePassive(NodeID nodeID) {
 //  a double ack locally but that is ok.  acknowledge is loose and can tolerate it. 
     if (activated) {
 //  remove the passive node from the sender first.  nothing else is going out
-      this.replicationSender.removePassive(nodeID);
+      BarrierCompletion barrierCompletion = new BarrierCompletion();
+      this.replicationSenderMessageSink.addSingleThreaded(new RemovePassiveMessage(nodeID, () -> barrierCompletion.complete()));
+      barrierCompletion.waitForCompletion();
       removeWaiters(nodeID);
```
myronkscott: Can passives be added and removed via synchronization?  They don't happen often and if we can remove the extra step and be more straightforward, I think it is better.

dso-l2/src/main/java/com/tc/objectserver/handler/ReplicationSenderMessage.java
```diff
@@ -0,0 +1,15 @@
+package com.tc.objectserver.handler;
```
myronkscott: I wonder if we can get rid of this new class and just use SyncActivityMessage as the type then handle flushes another way.

PR #738, opened 2017-10-06
TerracottaConnectionService: Throw exception when port in the given U…

ramsai1729:
…RI is malformed

### DIFF
--- a/connection-loader/src/main/java/com/terracotta/connection/api/TerracottaConnectionService.java
+++ b/connection-loader/src/main/java/com/terracotta/connection/api/TerracottaConnectionService.java
@@ -44,7 +44,7 @@ import java.util.concurrent.TimeoutException;
  * This is possible because the underlying connection knows how to probe the stripe for active nodes on start-up or
  * fail-over.
  */
-public class TerracottaConnectionService implements ConnectionService {
+public class TerracottaConnectionService extends AbstractConnectionService {
   private static final String SCHEME = "terracotta";
 
   private final EndpointConnector endpointConnector;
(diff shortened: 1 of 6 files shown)


myronkscott:
Please daggy to 5.3 and master

connection-loader/src/main/java/com/terracotta/connection/api/TerracottaConnectionService.java
```diff
@@ -75,11 +75,22 @@ public Connection connect(URI uri, Properties properties) throws ConnectionExcep
     TerracottaClientConfigParams clientConfig = new TerracottaClientConfigParams();
     String[] hosts = uri.getSchemeSpecificPart().split(",");
     for(String host : hosts) {
+
+      String[] tokens = host.split(":");
+      if(tokens.length >= 2) {
```
myronkscott: Probably should throw an exception if greater than 2.  I wonder if this will mess up with ipv6

ramsai1729: okay, will make changes to throw an exception. this change will not work with ipv6

ljacomet:
Had a chat with @ramsai1729 on this. I believe we should do something that would remain valid with IPv6, which what is done now is not.
However I confirm it makes a lot of sense to me to perform some level of validation on the text that is being parsed in this context.

myronkscott:
IPv6 is a problem but also similar connection services (diagnostic port, etc) do not get the benefit of this change.  I think we need to push this change higher or lower in the stack so we can write the validation once and have it available to all connection services.

ramsai1729:
I have refactored existing connection service impl code to share the uri parsing code and validating code and future implementations can use this code as well. This PR uses URI.parseServerAuthority() for validating the given host:port(s) instead of parsing and validating explicitly in our code. @myronkscott Please review and let me know if any changes needed.

PR #754, opened 2017-10-18
Direct Execution

myronkscott:
Add direct single threaded execution

Remove Hydrate Stages

Cleanup stage names


common/src/main/java/com/tc/async/impl/DirectSink.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+  }
+  
+  public static boolean isActivated() {
+    return ACTIVATED.get() == Thread.currentThread();
+  }
+  
+  private boolean isSingleThreaded() {
+    Thread thread = ACTIVATED.get();
+    return thread != null && thread == Thread.currentThread()
```
cschanck: isn't this the same as isActivated()?

cschanck: wait, isActivated() && isIdle.get()

common/src/main/java/com/tc/async/impl/StageImpl.java
```diff
@@ -32,6 +32,7 @@
 import com.tc.properties.TCPropertiesImpl;
 import com.tc.util.concurrent.QueueFactory;
 import com.tc.util.concurrent.ThreadUtil;
+import java.util.Arrays;
 
 /**
  * The SEDA Stage
```
cschanck: Evergreen comment.

PR #757, opened 2017-10-19
Passive sync changes

myronkscott:
add async passive channel
move syncing entity to its own pipeline
cleanup


myronkscott:
Closing to combine with major api changes

PR #758, opened 2017-10-26
Setting JAVA_HOME via installer setenv if available

akomakom:
Similar behavior to tc 4.x - if installed via installer, use the installer JVM


akomakom:
I meant to mention here also (in addition to internal jira) that this is not ready to merge... oh well, will need changes anyway.

PR #762, opened 2017-11-02
Fix #761 - Update retirement manager to use identity keys

### LINKED ISSUE #761: RetirementManager should use identity for keys
currently, retirement manager tracks retirement based on EntityMessage which has no hashCode or equals contract. Use a wrapper object to use object identity for keys.

myronkscott:


dso-l2/src/main/java/com/tc/objectserver/handler/RetirementManager.java
```diff
@@ -49,59 +49,57 @@
  * message, per key.
  */
 public class RetirementManager {
-  private final Map<EntityMessage, LogicalSequence> currentlyRunning;
-  private final Map<EntityMessage, LogicalSequence> waitingForDeferredRegistration;
+  private final Map<Identity<EntityMessage>, LogicalSequence> currentlyRunning;
```
ramsai1729: how about using https://docs.oracle.com/javase/8/docs/api/java/util/IdentityHashMap.html?

dso-l2/src/main/java/com/tc/objectserver/handler/RetirementManager.java
```diff
@@ -163,8 +157,9 @@ private void traverseDependencyGraph(List<Retiree> toRetire, LogicalSequence com
     }
   }
 
-  public synchronized void deferRetirement(EntityMessage invokeMessageToDefer, EntityMessage laterMessage) {
+  public void deferRetirement(EntityMessage invokeMessageToDefer, EntityMessage laterMessage) {
```
ramsai1729: could you explain why synchronized is not needed here?

dso-l2/src/main/java/com/tc/objectserver/handler/RetirementManager.java
```diff
@@ -190,16 +185,16 @@ public synchronized void entityWasDestroyed() {
     Assert.assertTrue(this.waitingForDeferredRegistration.isEmpty());
   }
   
-  public Map<String, Object> getState() {
+  public synchronized Map<String, Object> getState() {
```
ramsai1729: doesn't this block request processor threads?

dso-l2/src/main/java/com/tc/objectserver/handler/RetirementManager.java
myronkscott: that is a mistake will fix.

dso-l2/src/main/java/com/tc/objectserver/handler/RetirementManager.java
myronkscott: yes but the alternative is a random CME.  We should look into refactoring this class to be more concurrent at some point.

PR #770, opened 2017-11-14
Validate connection URIs

ChrisGreenaway:


connection-loader/src/main/java/com/terracotta/connection/api/AbstractConnectionService.java
```diff
@@ -48,6 +50,23 @@ public final Connection connect(URI uri, Properties properties) throws Connectio
     return internalConnect(clientConfig);
   }
 
+  private void validateURI(URI uri) throws ConnectionException {
+    String path = uri.getPath();
+    if (path != null && path.length() > 0 && !path.equals("/")) {
```
henri-tremblay: The path can't be null in a URI. And aren't you looking for `path.contains('/')`? I'm not sure what this code is for. Are we validating an URI like `terracotta://example.com:9540/my-application`?

connection-loader/src/main/java/com/terracotta/connection/api/AbstractConnectionService.java
```diff
@@ -48,6 +50,23 @@ public final Connection connect(URI uri, Properties properties) throws Connectio
     return internalConnect(clientConfig);
   }
 
+  private void validateURI(URI uri) throws ConnectionException {
```
henri-tremblay: Can be static and in fact could be in some util class. I would love to see unit tests on that.

myronkscott:
There is no issue open against this PR.  What is the purpose of this change and why is it needed?  I’m wondering if this is the right place to do this as it will break backwards compatibility.

connection-loader/src/main/java/com/terracotta/connection/api/AbstractConnectionService.java
ChrisGreenaway: I can do that.

connection-loader/src/main/java/com/terracotta/connection/api/AbstractConnectionService.java
ChrisGreenaway: A path can be null (see https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#getPath-- ).

We are trying to ensure that no-one specifies a path. So "terracotta://example.com:9540" and "terracotta://example.com:9540/" are valid - although the second one is dubious if you ask me.

henri-tremblay: Stupid me... I thought the path was the full URL. Now I get it. In fact you don't need the length then no? Could be `if(path != null && !path.equals("/"))`

connection-loader/src/main/java/com/terracotta/connection/api/URIUtils.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2011-2017 Software AG, Darmstadt, Germany and/or Software AG USA Inc., Reston, VA, USA, and/or its subsidiaries and/or its affiliates and/or their licensors.
+ * Use, reproduction, transfer, publication or disclosure is prohibited except as specifically provided for in your License Agreement with Software AG.
+ */
+package com.terracotta.connection.api;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+class URIUtils {
```
henri-tremblay: Please do a final class and private constructor

connection-loader/src/test/java/com/terracotta/connection/api/URIUtilsTest.java
```diff
@@ -0,0 +1,66 @@
...
+ * Use, reproduction, transfer, publication or disclosure is prohibited except as specifically provided for in your License Agreement with Software AG.
+ */
+package com.terracotta.connection.api;
+
+import org.junit.Test;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+public class URIUtilsTest {
```
henri-tremblay: In fact, the is a problem here. You should use a `ExpectedException` rule instead. Because right now we don't know if it's `new URI` or `validateTerracottaURI` which causes the exception. But if you do

```
URI uri = new URI("...");
expectedException.expect(URISyntaxException.class); // and because I'm crazy, I would also expect the message
URIUtils.validateTerracottaURI(uri);
```

It will fail when wanted only

connection-loader/src/main/java/com/terracotta/connection/api/AbstractConnectionService.java
mathieucarbou: @ChrisGreenaway : URIs such as `news:terracotta:9410` or `mailto:toto@terracotta:9410` will pass the validation and have null path. Is it normal ?

henri-tremblay: @mathieucarbou I don't think we should go crazy about it. The intent is, I think, so prevent paths, not to entirely validate the URI.

connection-loader/src/test/java/com/terracotta/connection/api/URIUtilsTest.java
chrisdennis: I agree except that stylistically I prefer using try catch constructs instead of ExpectedException rules.  To each his own.

henri-tremblay: Ouch... So sad and error prone.

```
URI uri = new URI("...");
try {
  URIUtils.validateTerracottaURI(uri);
  fail("Should throw an exception"); // error prone because forgotten
} catch(URISyntaxException e) {
 assertThat(e.getMessage()).isEqualsTo("...");
}
```
I've fix some errors like that in another OSS framework you know well.

ChrisGreenaway: Maybe - except that, at a system level, I don't particularly need / want to distinguish between strings rejected by new URI() and validateTerracottaURI().

However I don't mind changing the test to be more explicit, if you will feel happier.

ChrisGreenaway:
After some discussion with Myron on Friday, I've created this PR in terracotta-apis instead:
https://github.com/Terracotta-OSS/terracotta-apis/pull/283

connection-loader/src/test/java/com/terracotta/connection/api/URIUtilsTest.java
henri-tremblay: Yes, it will make me happier :-)

myronkscott:
Closed in favor of https://github.com/Terracotta-OSS/terracotta-apis/pull/283

Issue #781, opened 2017-11-22
Do not replay passive's monitoring tree after the server did become active


mathieucarbou:
There is an inconsistency which is failing when we activate the assertions. The newly active server is receiving some `addNode()` calls with data related to the old passive server _after_ it became active.

Here are the logs:

1. `IStripeMonitoring.serverDidBecomeActive` is called on consumerId 0 (platform)
2. the active server is rebuilding the tree (`addNode()` calls on consumer id 0) like `[0] addNode(testServer1, platform, clients)`
3. at the end, we receive some calls on the channel used to send the addNode() and data calls from passive to active:

```
IStripeMonitoringDataListenerAdapter - [2] addNode(testServer1, , registry)
DefaultDataListener - [2] setState(2, testServer1, [registry])
```

These entries are not required to be kept and replayed because the monitoring tree is rebuilding already and we don't need the old data. Also, I cannot ignore these replays in the test (assert == true) because we could miss some m&m bugs (in example a statistic collector that is still running in background on the passive entity's consumerId).

So I suggest we just drop the tree after a failover. In a more general way, I think platform does not need to keep a copy of the monitoring tree. If some persistence are needed at one point and a replay is needed, perhaps it is up to the implementor to decide that ?

In our case, this replay is not needed and we have to deal with it, and is really hard to just  ignore.

```
2017-11-22 15:01:29,945 [WorkerThread(l2_state_change_stage, 0)] TRACE org.terracotta.management.service.monitoring.IStripeMonitoringPlatformListenerAdapter - [0] serverDidBecomeActive(testServer1)
2017-11-22 15:01:29,945 [WorkerThread(l2_state_change_stage, 0)] TRACE org.terracotta.management.service.monitoring.TopologyService - [0] serverDidBecomeActive(testServer1)
2017-11-22 15:01:29,948 [WorkerThread(l2_state_change_stage, 0)] TRACE org.terracotta.management.service.monitoring.DefaultClientMonitoringService - [1] onBecomeActive()
2017-11-22 15:01:29,948 [WorkerThread(l2_state_change_stage, 0)] TRACE org.terracotta.management.service.monitoring.DefaultManagementService - [2] onBecomeActive()
[...]

PR #783, opened 2017-11-23
[PROPOSAL] Close #782: Do not flush cached data sent with pushBestEffortsData() after promotion

### LINKED ISSUE #782: Do not replay (flush) cached data from pushBestEffortsData after promotion
After a promotion, `BestEffortsMonitoring.flushAfterActivePromotion(BestEffortsMonitoring.java:77)` is called. 
This as the effect of sending in the monitoring service some data that were created in a passive server related to some passive entities that do not exist anymore.
The goal of `pushBestEffortsData` is to be non reliable. I don't think data should be cached and replayed after a promotion, especially because 1) this is a non reliable channel and 2) data is constructed related to an old context where the server was passive and had passive entities. So this data is now completely obsolete and should be discarded.

mathieucarbou:
For context, see discussions in PR and issues:

- https://github.com/Terracotta-OSS/terracotta-platform/pull/413
- https://github.com/Terracotta-OSS/terracotta-core/issues/781
- https://github.com/Terracotta-OSS/terracotta-core/issues/782

With this changes, PR https://github.com/Terracotta-OSS/terracotta-platform/pull/413 will pass.

Data cached from `pushBestEffortsData ` calls is cleared and not replayed after a promotion.

This change will fix https://github.com/Terracotta-OSS/terracotta-core/issues/781, https://github.com/Terracotta-OSS/terracotta-platform/pull/413 and will cancel issue https://github.com/Terracotta-OSS/terracotta-core/issues/781.

This will also let the `IMonitoringProducer.addNode()` method unused from any M&M classes. We will only use `IMonitoringProducer.pushBestEffortsData()`

### DIFF
--- a/dso-l2/src/main/java/com/tc/services/BestEffortsMonitoring.java
+++ b/dso-l2/src/main/java/com/tc/services/BestEffortsMonitoring.java
@@ -60,23 +60,7 @@ public class BestEffortsMonitoring {
   public synchronized void flushAfterActivePromotion(PlatformServer thisServer, TerracottaServiceProviderRegistry globalRegistry) {
     // We no longer care about the timer so clear it, if one exists.
     ensureTimerCancelled();
-    
-    // Walk each consumerID, looking up their registries, and flushing all entries to the implementation.
-    for (Map.Entry<Long, Map<String, Serializable>> perConsumerEntry : this.bestEffortsCache.entrySet()) {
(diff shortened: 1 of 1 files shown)


mathieucarbou:
talked with @myronkscott : a solution will be implemented in PR https://github.com/Terracotta-OSS/terracotta-platform/pull/413/commits

PR #785, opened 2017-11-24
[master] Cleaning up for Java 9 support

ljacomet:
* Removed obsolete version parsing code
* Removed usage of `sun.misc.BASE64Encode`
  * Which led to the removal of all connection security related code inherited from 4.x


dso-l1/src/main/java/com/tc/object/DistributedObjectClient.java
```diff
@@ -428,14 +416,8 @@ private void openChannel() throws InterruptedException {
...
-            Assert.assertNotNull(securityManager);
-            username = config.getSecurityInfo().getUsername();
-            pw = securityManager.getPasswordForTC(config.getSecurityInfo().getUsername(), hostname, port);
-          } else {
-            pw = null;
-            username = null;
-          }
+          pw = null;
+          username = null;
           this.channel.open(infos, username, pw);
```
ramsai1729: can't we use this.channel.open(infos) here?

ljacomet: indeed, fixing

ljacomet:
@ramsai1729 I have addressed the comment - can you look again?

Issue #794, opened 2017-12-05
Issues running under the latest IBM JDK 1.8


ljacomet:
SR 5 of the IBM JDK 1.8 causes the server to fail to start.

```
java.lang.AssertionError: Old or Tenured Memory pool Not found : [class storage, miscellaneous non-heap storage, JIT code cache, JIT data cache, tenured-SOA, tenured-LOA, nursery-allocate, nursery-survivor]
	at com.tc.runtime.TCMemoryManagerJdk15PoolMonitor.getOldGenMemoryPoolBean(TCMemoryManagerJdk15PoolMonitor.java:60)
	at com.tc.runtime.TCMemoryManagerJdk15PoolMonitor.<init>(TCMemoryManagerJdk15PoolMonitor.java:45)
	at com.tc.runtime.TCRuntime.init(TCRuntime.java:47)
	at com.tc.runtime.TCRuntime.<clinit>(TCRuntime.java:32)
	at com.tc.management.beans.TCServerInfo.<init>(TCServerInfo.java:86)
	at com.tc.server.TCServerImpl.registerServerMBeans(TCServerImpl.java:422)
	at com.tc.server.TCServerImpl.createDistributedObjectServer(TCServerImpl.java:397)
	at com.tc.server.TCServerImpl.startDSOServer(TCServerImpl.java:387)
	at com.tc.server.TCServerImpl.access$300(TCServerImpl.java:73)
	at com.tc.server.TCServerImpl$StartAction.execute(TCServerImpl.java:365)
	at com.tc.lang.StartupHelper.startUp(StartupHelper.java:52)
	at com.tc.server.TCServerImpl.startServer(TCServerImpl.java:382)
	at com.tc.server.TCServerImpl.start(TCServerImpl.java:203)
	at com.tc.server.TCServerMain.main(TCServerMain.java:74)
```

PR #795, opened 2017-12-05
[release/5.3] Issue #794 Degrade when memory pools not identified

ljacomet:
Also update the max direct memory value used on command line, IBM VM no
longer parses 2^63-1 properly

### DIFF
--- a/terracotta-kit/src/assemble/server/bin/start-tc-server.bat
+++ b/terracotta-kit/src/assemble/server/bin/start-tc-server.bat
@@ -48,11 +48,11 @@ if not defined JAVA_HOME (
 )
 
 set JAVA_HOME="%JAVA_HOME:"=%"
-for %%C in ("\bin\java -d64 -server -XX:MaxDirectMemorySize=9223372036854775807" ^
-			"\bin\java -server -XX:MaxDirectMemorySize=9223372036854775807" ^
-			"\bin\java -d64 -client  -XX:MaxDirectMemorySize=9223372036854775807" ^
-			"\bin\java -client -XX:MaxDirectMemorySize=9223372036854775807" ^
-			"\bin\java -XX:MaxDirectMemorySize=9223372036854775807") ^
+for %%C in ("\bin\java -d64 -server -XX:MaxDirectMemorySize=1048576g" ^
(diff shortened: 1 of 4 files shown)


terracotta-kit/src/assemble/server/bin/start-tc-server.bat
```diff
@@ -48,11 +48,11 @@ if not defined JAVA_HOME (
...
-			"\bin\java -server -XX:MaxDirectMemorySize=9223372036854775807" ^
-			"\bin\java -d64 -client  -XX:MaxDirectMemorySize=9223372036854775807" ^
-			"\bin\java -client -XX:MaxDirectMemorySize=9223372036854775807" ^
-			"\bin\java -XX:MaxDirectMemorySize=9223372036854775807") ^
+for %%C in ("\bin\java -d64 -server -XX:MaxDirectMemorySize=1048576g" ^
+			"\bin\java -server -XX:MaxDirectMemorySize=1048576g" ^
+			"\bin\java -d64 -client  -XX:MaxDirectMemorySize=1048576g" ^
+			"\bin\java -client -XX:MaxDirectMemorySize=1048576g" ^
+			"\bin\java -XX:MaxDirectMemorySize=1048576g") ^
 do (
```
myronkscott: Maybe we try both values in the for loop?  I'm worried about other platforms not handling the new style.

ljacomet: This has been tested with an Oracle JVM and the IBM one. Which other are you thinking of?

lorban: `Long.MAX_VALUE - 1` just proved to be problematic so I definitely wouldn't rely on it anymore. Having spent a few minutes in the Hotspot initialization code that slices up the native heap, it already is almost a miracle that this value ever worked at all.

This format is standard for `-Xmx` so I'm tempted to believe the same code will always be used to parse the `XX` memory size flags as well. We could convert the 1 PB @ljacomet used here from gigabytes down to bytes but that sounds to me like splitting hair.

lorban: For future reference: http://mail.openjdk.java.net/pipermail/hotspot-runtime-dev/2017-April/023124.html

PR #806, opened 2017-12-14
Improve exception message when connection to TSA server is not establ…

chaktatha:
…ished because either it is not running or a wrong port number is given in client side.


terracotta-org:

dso-l1/src/main/java/com/tc/object/DistributedObjectClient.java
```diff
@@ -697,7 +709,7 @@ private void shutdown(boolean fromShutdownHook, boolean forceImmediate) {
         clientStopped.notifyAll();
       }
       if (this.channel != null && !this.channel.getProductId().isInternal()) {
-        DSO_LOGGER.info("closing down Terracotta Connection hook=" + fromShutdownHook + " force=" + forceImmediate + " channel=" + this.channel.getChannelID() + " client=" + this.channel.getClientID());
+        DSO_LOGGER.debug("closing down Terracotta Connection hook=" + fromShutdownHook + " force=" + forceImmediate + " channel=" + this.channel.getChannelID() + " client=" + this.channel.getClientID());
       }
```
myronkscott: I believe this to be important information worthy of INFO level logging.  What is the rationale behind this change?

dso-l1/src/main/java/com/tc/object/DistributedObjectClient.java
```diff
@@ -445,9 +455,11 @@ private void waitForHandshake() {
     this.clientHandshakeManager.waitForHandshake();
     if (this.channel != null) {
       final TCSocketAddress remoteAddress = this.channel.getRemoteAddress();
-      final String infoMsg = "Connection successfully established to server at " + remoteAddress;
-      if (!this.channel.getProductId().isInternal()) {
-        DSO_LOGGER.info(infoMsg);
+      if(remoteAddress!=null){
```
myronkscott: If remote address is null, there is probably a larger problem at hand.  What is driving this change?

dso-l1/src/main/java/com/tc/object/DistributedObjectClientFactory.java
```diff
@@ -94,8 +95,13 @@ public Void call() throws Exception {
       client.start();
       String timeout = properties.getProperty(ConnectionPropertyNames.CONNECTION_TIMEOUT, "0");
       if (!client.waitForConnection(Long.parseLong(timeout), TimeUnit.MILLISECONDS)) {
-//  timed out, shutdown the extra threads and return null;
+//  timed out, shutdown the extra threads and return null if internal digested exception is null;
+//  otherwise throw TimeoutException wrapping the original exception
         client.shutdown();
+        Exception digestedException = client.getLastDigestedException();
```
myronkscott: This change is confusing to me.  I understand some more information is desired about why the connection timed out but I don't think the method actually provides that as the connection is likely trying multiple servers for a connection.  Can you please open a GitHub issue with some context so we can throughly review this PR.

myronkscott: I don't like this style.  We should not add the TimeoutException here.  The last exception thrown on the last connection tried can be just as confusing as nothing at all.  In my opinion, if you want exception information, continuing to return null here is better.  If an upper layer would like to inquire about connection attempts and history, I think a map of each connection attempt with the result would be more appropriate.  You could add an accessor method to DOClient such as getConnectionHistory() to access the map.

dso-l1/src/main/java/com/tc/object/DistributedObjectClient.java
```diff
@@ -153,6 +154,8 @@
   private final SetOnceFlag                          clientStopped                       = new SetOnceFlag();
   private final SetOnceFlag                          connectionMade                       = new SetOnceFlag();
   private final SetOnceRef<Exception>                exceptionMade                       = new SetOnceRef<Exception>();
+
+  private final AtomicReference<Exception>           lastDigestedExceptionMade               = new AtomicReference<>();
```
myronkscott: Use a collection of destination and result, possibly timestamp as well.

dso-l1/src/main/java/com/tc/object/DistributedObjectClient.java
```diff
@@ -399,6 +402,10 @@ public boolean waitForConnection(long timeout, TimeUnit units) throws Interrupte
     return connectionMade.isSet();
   }
 
+  public Exception getLastDigestedException(){
```
myronkscott: Change to something that returns results for attempts to each target destination.

dso-l1/src/main/java/com/tc/object/DistributedObjectClientFactory.java
chaktatha: I have a query here. Would it really be helpful for upper layer to go for an additional method invocation to get the root cause? Would it not be an additional overhead? Typically, exception should contain enough information indicating the root cause. If the last exception thrown is included inside the TimeoutException, client might have some idea about the root cause from the exception itself. Please let me know your thoughts. If you still think including last exception (on the last connection tried) does not make sense and can be confusing, please let me know. I will go ahead and change the implementation as you have suggested.

I am accepting other comments (i.e. fixing the formatting and bringing back the INFO level). I will make changes as suggested.

myronkscott: Consider a stripe with two nodes.  The user puts both nodes in the uri.  The hostname of the first node is spelled wrong, the second correctly.  Now the first node is up but unreachable because of the incorrect hostname, the second node is down.  The last exception happens to be on the second node (down).  Now the user, going by only the stack trace assumes the cluster is down when it is actually up and running.  So the reason for the timeout exception in this case is the mis-spelled hostname and not what was reported by the timeout exception.  If we are going to report these exceptions at the code level, we should provide the full picture (per-connection failure) rather than just the very last thing that has happened.

myronkscott: Another approach to this might be to add an exception listener interface that could be registered so that each underlying exception could be delivered

myronkscott:
#847 merged instead

PR #811, opened 2017-12-21
Client builder refactoring to enable service-loader based lookup

saurabhagas:


terracotta-org:

PR #823, opened 2018-01-05
i-822 Add support for logging extra message after logging version

rkavanap:


dso-l2/src/main/java/com/tc/server/TCServerMain.java
```diff
@@ -100,6 +100,11 @@ private static void writeVersion() {
       String longPatchString = info.toLongPatchString();
       LOGGER.info(longPatchString);
     }
+
+    String versionMessage = info.versionMessage();
+    if (versionMessage != null && versionMessage.length() > 0) {
```
ljacomet: This will never be null by contract and I would replace the length check by `String.isEmpty()`

PR #826, opened 2018-01-08
Misc. Refactorings

myronkscott:
See commit messages

Refactor Sink interface
Add hydrate stage
Add more visibility front half (networking) stages


common/src/main/java/com/tc/async/impl/PipelineMonitor.java
```diff
@@ -102,6 +106,21 @@ public String toString() {
     StringBuilder builder = new StringBuilder("\n" + parent + "\n");
     items.forEach(r->builder.append("\t" + r + "\n"));
     builder.append("time:" + TimeUnit.NANOSECONDS.toMicros(items.get(items.size()-1).time - items.get(0).time));
+    builder.append('\n');
+    Set<String> names = items.stream().map(i->i.name).collect(Collectors.toSet());
```
cschanck: fancy

common/src/main/java/com/tc/async/api/Sink.java
```diff
@@ -28,18 +28,5 @@
...
-   * 
-   * TODO:  Split out the multi-threaded cases to avoid the casts required in the implementations.  This hack is just a
-   * stop-gap to keep the change to add types to SEDA smaller.
-   * 
-   * @param context
-   */
-  public void addMultiThreaded(EC context);
-
-  public void close();
+  public void addToSink(EC context);
```
cschanck: Javadoc?

PR #830, opened 2018-01-19
protect against client-server version/protocol mismatch

AlbertGevorgyan:
* Write info message to client log if versions are different but no
error is thrown.
* Added unit tests, upgraded powermock to 1.6.1.


AlbertGevorgyan:
Have refactored ClientHandshakeManagerTest for better readability.

PR #835, opened 2018-01-25
Terracotta voter scripts and kitting

albinsuresh:


terracotta-kit/src/assemble/client/logging/impl/logback.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
albinsuresh: This was added for the standalone voter logging.

PR #836, opened 2018-01-30
Small Changes

myronkscott:
Remove SSL special handling
Improve consistency logging
Fix state reporting for TCConnection
Remove unused SSL handling


voter/src/main/java/org/terracotta/voter/ActiveVoter.java
```diff
@@ -143,6 +144,7 @@ public void registerAndHeartbeat(ExecutorService executorService, ClientVoterMan
                   long result = voterManager.vote(id, election);
                   LOGGER.info("own the vote, voting {} {} {}", voterManager.getTargetHostPort(), election, result);
                 } else if (owner.isConnected()) {
+                  owner = voteOwner.get();
```
albinsuresh: Unused? Or is there a logic here that I didn't understand?

voter/src/main/java/org/terracotta/voter/ActiveVoter.java
```diff
@@ -209,6 +215,22 @@ public void registerAndHeartbeat(ExecutorService executorService, ClientVoterMan
     LOGGER.warn("Rejected by all servers. Attempting to re-register");
   }
   
+  private static void sleepFor10() {
+    try {
+      TimeUnit.SECONDS.sleep(10);
+    } catch (InterruptedException ie) {
+      //
```
albinsuresh: Not sure if this should be ignored especially since we need to support interruption in `deregister`.

voter/src/main/java/org/terracotta/voter/ActiveVoter.java
myronkscott: the owner of the vote may have changed if you have 3 threads running for 3+ servers.  This just resets the owner for the next round.

voter/src/main/java/org/terracotta/voter/ActiveVoter.java
myronkscott: will reset interrupt.

common/src/main/java/com/tc/net/core/CoreNIOServices.java
```diff
@@ -316,8 +316,8 @@ public String getName() {
     Map<String, Object> state = new LinkedHashMap<>();
     state.put("name", this.commThreadName);
     state.put("weights", this.clientWeights);
-    state.put("writer", this.readerComm.getCommState());
-    state.put("reader", this.writerComm.getCommState());
+    state.put("writer", this.writerComm.getCommState());
+    state.put("reader", this.readerComm.getCommState());
```
cschanck: ?

myronkscott: mis-labeled. needed to reverse.

cschanck: ah

PR #843, opened 2018-02-07
avoid string append if logging is not needed

myronkscott:


henri-tremblay:
Why not using placeholders? `logger.debug("Scheduling {} on {}: {}",  next.request.getAction(), getID(), getConsumerID());`

myronkscott:
Yeah that is probably better. Will keep that in mind for the next time around.

Issue #845, opened 2018-02-08
Proper exception for already shutdown connection


AbfrmBlr:

myronkscott:
#846

PR #846, opened 2018-02-08
New exception for shutdown connections #845

AbfrmBlr:
This is the only place I think we need, but I may be wrong.
So I think as and when we find some more places, we replace Illegalstates with `ConnectionShutdownException` or `ConnectionClosedException`, which ever fits/suits


AbfrmBlr:
Good point. But that would mean moving it to connection api, (though it will not change the api).
@myronkscott , @cschanck wdyt?

AbfrmBlr:
https://github.com/Terracotta-OSS/terracotta-apis/pull/289
I have opened this PR in apis, it only adds the exception

AbfrmBlr:
test this please

AbfrmBlr:
test this please

PR #851, opened 2018-02-14
 Avoid server configuration parsing by voter

albinsuresh:
Reviewing the first commit in isolation would be better as the second one is all about text replaces.


voter/src/main/java/org/terracotta/voter/TCConfigParserUtil.java
```diff
@@ -0,0 +1,51 @@
...
+
+    NodeList serverList = document.getElementsByTagName(SERVER_NODE);
+    String[] hostPorts = new String[serverList.getLength()];
+
+    for (int i = 0; i < serverList.getLength(); i++) {
+      Node node = serverList.item(i);
+      if (node.getNodeType() == Node.ELEMENT_NODE) {
+        Element serverNode = (Element) node;
+        String hostname = serverNode.getAttributes().getNamedItem(HOST_ATTR).getNodeValue();
+        int port = Integer.parseInt(serverNode.getElementsByTagName(TSA_PORT_NODE).item(0).getTextContent());
```
ramsai1729: what if <tsa-port> is not present?

myronkscott: good point.  You probably need to implement default port

myronkscott:
Closing as merged in #853

PR #852, opened 2018-02-14
Allow to unset a TCProperty

henri-tremblay:


myronkscott:
Why do you need this?  There are a bunch of assumptions in the codebase that TCProperties are well known and have defaults.  Allowing arbitrary unsetting seems unwise.

henri-tremblay:
@myronkscott `TCPropertiesConsts.CLIENT_MAX_PENDING_REQUESTS` has no default. So after setting it, I need to unset it.

PR #856, opened 2018-02-15
Prevent re-registration of an existing voter when voting in progress

albinsuresh:


dso-l2/src/main/java/com/tc/l2/state/ServerVoterManagerImpl.java
```diff
@@ -67,7 +67,7 @@ public int getVoterLimit() {
 
   @Override
   public synchronized long registerVoter(String id) {
-    if (voters.containsKey(id)) {
+    if (!votingInProgress && voters.containsKey(id)) {
       //  already registered.  double register is not supported
```
myronkscott: I don't understand this change.  HEARTBEAT_RESPONSE means that the voter is already registered.  I don't think timeout needs to be considered but maybe.

albinsuresh:
Fixed by other means in #859

PR #857, opened 2018-02-15
reverting 25b5063

myronkscott:


myronkscott:
reverting due to improper assumption that all server to client messages are coming from the pipeline.

PR #878, opened 2018-03-02
Fix #877 - Do not remove waiters on GroupException

### LINKED ISSUE #877: GroupExceptions should not clear waiters
waiters should not be cleared by GroupExceptions.  The waiters should always be cleared by passive removal.

myronkscott:


myronkscott:
test this please

PR #882, opened 2018-03-06
Redirects are only advisory not handshake error. 

myronkscott:
Diagnostic clients must use this connection and cannot be blocked.


myronkscott:
I'm working on a test but this area of code is very hard to hit with a unit test.  Verified currently with an integration test

galvan-support/src/test/java/org/terracotta/testing/rules/BasicExternalClusterFOPConsistencyVoterIT.java
```diff
@@ -31,7 +31,6 @@
   @ClassRule
   public static final Cluster CLUSTER = BasicExternalClusterBuilder.newCluster(2).withFailoverPriorityVoterCount(1).build();
 
-  @Test(expected = TimeoutException.class)
```
albinsuresh: Removed the @Test annotation as well by mistake.

albinsuresh:
test this please

PR #893, opened 2018-04-11
Enable compatibility across minor versions.

AlbertGevorgyan:
Enabled only compatibility within a major version - not clear whether we need it across major versions as well? Also cleaned up warnings in test code, and switched VersionCompatibilityTest from junit.framework to org.junit.

### DIFF
--- a/common/src/test/java/com/tc/util/version/VersionCompatibilityTest.java
+++ b/common/src/test/java/com/tc/util/version/VersionCompatibilityTest.java
@@ -18,17 +18,17 @@
  */
 package com.tc.util.version;
 
-import junit.framework.TestCase;
+import org.junit.Test;
 
-public class VersionCompatibilityTest extends TestCase {
+import static com.tc.util.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.fail;
 
-  private VersionCompatibility versionCompatibility;
+public class VersionCompatibilityTest {
 
-  @Override
-  protected void setUp() throws Exception {
-    this.versionCompatibility = new VersionCompatibility();
-  }
+  private final VersionCompatibility versionCompatibility = new VersionCompatibility();
 
+  @Test
   public void testNull() {
     try {
       versionCompatibility.isCompatibleClientServer(v("1.0.0"), null);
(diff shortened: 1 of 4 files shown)


common/src/test/java/com/tc/util/version/VersionCompatibilityTest.java
```diff
@@ -18,17 +18,17 @@
...
+import static org.junit.Assert.fail;
 
-  private VersionCompatibility versionCompatibility;
+public class VersionCompatibilityTest {
 
-  @Override
-  protected void setUp() throws Exception {
-    this.versionCompatibility = new VersionCompatibility();
-  }
+  private final VersionCompatibility versionCompatibility = new VersionCompatibility();
```
AlbertGevorgyan: Is immutable so no need to re-create for each test.

dso-l2/src/test/java/com/tc/net/groups/TCGroupManagerImplTest.java
```diff
@@ -561,13 +561,13 @@ protected String getVersion() {
 
     // Incompatible version, higher weights. Close down the handshake, the other guy will zap himself.
     MessageChannel channel1 = mockMessageChannel();
-    tcGroupManager.receivedHandshake(mockHandshakeMessage(channel1, "4.3.1", new long [] { 4, 5 }));
+    tcGroupManager.receivedHandshake(mockHandshakeMessage(channel1, "5.3.1", new long [] { 4, 5 }));
     verify(channel1).close();
```
AlbertGevorgyan: This implies that VERSION_COMPATIBILITY_CHECK property is set to true. Setting it explicitly in the code would require initialising the static class TCPropertiesImpl by PowerMock (see ClientHandshakeManagerTest above), and then setting other properties explicitly as well, so decided to leave the implicit configuration as is. Changed the version difference from minor to major to get the desired failure.

AlbertGevorgyan:
We already have the TCPropertiesConsts.VERSION_COMPATIBILITY_CHECK property, so can switch off compatibility check entirely. The only change in behaviour in this PR is completely disabling checks for minor version. Only major will be checked by default.

dso-l2/src/test/java/com/tc/net/groups/TCGroupManagerImplTest.java
myronkscott: In order to set the default for the platform change this line to false

https://github.com/Terracotta-OSS/terracotta-core/blob/master/common/src/main/resources/com/tc/properties/tc.properties#L330

AlbertGevorgyan: Not sure we want to change the default yet? I think myself and Chris D agreed just to ensure that it is configurable.

myronkscott:
can you change to a daggy merge to both release/5.4 and master

PR #896, opened 2018-04-13
Disable version compatibility check.

AlbertGevorgyan:
Completely disabled version compatibility checks in code. Removed configuration. Left the VersionCompatibility class and its test as they are (can we deprecate?)

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/impl/ServerPersistenceVersionChecker.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/impl/ServerPersistenceVersionChecker.java
@@ -19,52 +19,30 @@
 
 package com.tc.objectserver.impl;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.tc.objectserver.persistence.ClusterStatePersistor;
 import com.tc.util.ProductInfo;
 import com.tc.util.version.Version;
-import com.tc.util.version.VersionCompatibility;
 
 /**
  * @author tim
  */
 public class ServerPersistenceVersionChecker {
-  private static final Logger LOGGER = LoggerFactory.getLogger(ServerPersistenceVersionChecker.class);
-  private final ClusterStatePersistor clusterStatePersistor;
   private final ProductInfo productInfo;
-  private final VersionCompatibility versionCompatibility;
 
-  public ServerPersistenceVersionChecker(ClusterStatePersistor clusterStatePersistor) {
-    this(clusterStatePersistor, ProductInfo.getInstance(), new VersionCompatibility());
+  ServerPersistenceVersionChecker() {
+    this(ProductInfo.getInstance());
   }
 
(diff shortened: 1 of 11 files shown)


dso-l2/src/main/java/com/tc/objectserver/impl/ServerPersistenceVersionChecker.java
```diff
@@ -47,24 +47,12 @@ public ServerPersistenceVersionChecker(ClusterStatePersistor clusterStatePersist
     this.versionCompatibility = versionCompatibility;
   }
 
-  private boolean checkVersion(Version persisted, Version current) {
-    if (persisted != null) { return versionCompatibility.isCompatibleServerPersistence(persisted, current); }
-    return true;
-  }
-
   public void checkAndSetVersion() {
```
myronkscott: Worth changing the name of this method to reflect what it is doing?

AlbertGevorgyan: I would rather refactor as well, as the method has an implicit side effect on a class field.

dso-l2/src/test/java/com/tc/objectserver/impl/ServerPersistenceVersionCheckerTest.java
```diff
@@ -45,66 +42,69 @@
...
     persistedVersion("1.0.0");
     verifyUpdatedTo("1.1.0");
   }
 
-  public void testMinorVersionDrop() throws Exception {
+  public void testMinorVersionDrop() {
     currentVersion("1.0.0");
     persistedVersion("1.1.0");
-    verifyExceptionAndNoUpdate();
+    verifyNoUpdate();
```
AlbertGevorgyan: Note change in behaviour: current < persistent is ignored.

dso-l2/src/main/java/com/tc/objectserver/impl/ServerPersistenceVersionChecker.java
```diff
@@ -19,52 +19,30 @@
...
-    this.versionCompatibility = versionCompatibility;
-  }
-
-  private boolean checkVersion(Version persisted, Version current) {
-    if (persisted != null) { return versionCompatibility.isCompatibleServerPersistence(persisted, current); }
-    return true;
   }
 
-  public void checkAndSetVersion() {
+  void checkAndBumpPersistedVersion(ClusterStatePersistor clusterStatePersistor) {
```
AlbertGevorgyan: Changed method name as suggested by Myron. Also refactored to make the side effect on clusterStatePersistor more explicit by passing it in as a method argument rather than a field.

myronkscott:
While all this is fine, I'm a little concerned about 4x <-> 10x server or persistence mixing.  Do you think there is something we can do to at least prevent attempted interop between those versions?

AlbertGevorgyan:
Can we hard code the minimal version that is not subject to compatibility checks? As an alternative, can we configure ranges of compatible versions?

PR #910, opened 2018-07-12
Add some simple Javadoc to BufferManager

ChrisGreenaway:


common/src/main/java/com/tc/net/core/BufferManager.java
```diff
@@ -27,12 +27,40 @@
...
+   * @return the number of bytes read
+   */
   int forwardToWriteBuffer(ByteBuffer src);
 
+  /**
+   * Called immediately after a call to a forwardToWriteBuffer() method. It will be called repeatedly until the cumulative
+   * total of bytes sent matches the number of bytes returned by the forwardToWriteBuffer() call.
+   * This method should write data to the channel
+   * @return the number of bytes sent
+   * @throws IOException
```
henri-tremblay: Why is the IOException thrown?

common/src/main/java/com/tc/net/core/BufferManager.java
```diff
@@ -27,12 +27,40 @@
...
+   * Called when there is data to write to the channel and the channel is accepting writes.
+   * It is called from the connection's write thread.
+   * This method should read data from the supplied ByteBuffer.
+   * @param src the ByteBuffer from which data should be read
+   * @return the number of bytes read
+   */
   int forwardToWriteBuffer(ByteBuffer src);
 
+  /**
+   * Called immediately after a call to a forwardToWriteBuffer() method. It will be called repeatedly until the cumulative
```
henri-tremblay: You can put a {@link #forwardToWriteBuffer()) for all the references in the javadoc.

common/src/main/java/com/tc/net/core/BufferManager.java
```diff
@@ -27,12 +27,40 @@
  * @author Ludovic Orban
  */
 public interface BufferManager {
+  /**
+   * Called immediately after a call to recvToBuffer(). It will be called again if it returns a non-zero value.
```
lorban: `It will be called again if it returns a non-zero value.` states an implementation detail of a consumer of this interface, namely the comm stack.

common/src/main/java/com/tc/net/core/BufferManager.java
```diff
@@ -27,12 +27,40 @@
...
+   * It is called from the connection's read thread.
+   * This method should write data into the supplied ByteBuffer.
+   * @param dest the ByteBuffer into which data should be written
+   * @return the number of bytes written
+   */
   int forwardFromReadBuffer(ByteBuffer dest);
 
+  /**
+   * Called when there is data to write to the channel and the channel is accepting writes.
+   * It is called from the connection's write thread.
```
lorban: `It is called from the connection's write thread.` is an implementation detail of the comm stack.

common/src/main/java/com/tc/net/core/BufferManager.java
```diff
@@ -27,12 +27,40 @@
...
+   * This method should read data from the supplied ByteBuffer.
+   * @param src the ByteBuffer from which data should be read
+   * @return the number of bytes read
+   */
   int forwardToWriteBuffer(ByteBuffer src);
 
+  /**
+   * Called immediately after a call to a forwardToWriteBuffer() method. It will be called repeatedly until the cumulative
+   * total of bytes sent matches the number of bytes returned by the forwardToWriteBuffer() call.
```
lorban: `It will be called repeatedly until the cumulative total of bytes sent matches the number of bytes returned by the forwardToWriteBuffer() call.` is again an implementation detail of the comm stack.

common/src/main/java/com/tc/net/core/BufferManager.java
```diff
@@ -27,12 +27,40 @@
...
+   * This method should write data to the channel
+   * @return the number of bytes sent
+   * @throws IOException
+   */
   int sendFromBuffer() throws IOException;
 
+  /**
+   * This method is called when the connection sees that data is available to read from the channel.
```
lorban: `This method is called when the connection sees that data is available to read from the channel.` is an implementation detail of the comm stack.

common/src/main/java/com/tc/net/core/BufferManager.java
```diff
@@ -27,12 +27,40 @@
...
+   * It is called from the connection's read thread.
```
lorban: `It is called from the connection's read thread.` is an implementation detail of the comm stack.

common/src/main/java/com/tc/net/core/BufferManager.java
```diff
@@ -27,12 +27,40 @@
...
+   * This method should read from the channel. If it does not then it will be called again.
```
lorban: `If it does not then it will be called again.` is an implementation detail of the comm stack.

common/src/main/java/com/tc/net/core/BufferManager.java
```diff
@@ -27,12 +27,40 @@
...
+   * @return the number of bytes available to be read using a forwardFromReadBuffer() method - although this value is currently ignored
```
lorban: `the number of bytes available to be read using a forwardFromReadBuffer() method` is wrong, it actually is the # of bytes that were received from the socket channel.

lorban: `although this value is currently ignored` is an implementation detail of the comm stack.

ChrisGreenaway:
1) @lorban suggests that we do not include anything that relates to comm stack implementation details, however these details are important when writing an implementation of BufferManager. For example, if you don't know that a particular method will always be called by the same thread then you have to worry about Java memory guarantees.
2) @henri-tremblay wants the Javadoc to have more information on when an IOException is thrown. I would suggest that when someone has something useful to write there, then they can write it. This PR is adding useful documentation and I have nothing useful to say about IOException.
3) @henri-tremblay wants a link to forwardToWriteBuffer() - but there are two methods with that name and I am referring to both of them.

PR #916, opened 2018-07-17
Make DiagnosticHandler a service

saurabhagas:


saurabhagas:
test this please

myronkscott:
These operations are defect interface of tc-core and should not be pluggable.

PR #919, opened 2018-07-18
Adapted to PlatformService interface changes and added implementation for TCServerImpl::stop

saurabhagas:
Added implementation for TCServerImpl::stop and make everyone use it instead of doing `Runtime.getRuntime().exit()`


saurabhagas:
test this please

PR #943, opened 2018-08-22
[daggy-5.4] Extended connection options for voter

albinsuresh:


voter/src/main/java/org/terracotta/voter/ClientVoterManagerImpl.java
```diff
@@ -58,11 +58,12 @@ public String getTargetHostPort() {
   }
 
   @Override
-  public void connect() {
+  public void connect(Optional<Properties> connectionProps) {
```
saurabhagas: I didn't see the reason why `Optional<Properties>` needs to be used everywhere instead of just `Properties`. `ConnectionFactory` and `ConnectionService` APIs take a `Properties` object, and so do the code places which call these APIs. Everywhere we just take a `Properties` object and pass it along to the connection creation. (I understand that this code predates the introduction of `Optional` in Java)

I'd have understood this if there was a need to make defensive copies of `Properties` to preserver the immutability of objects. Moreover, the `Properties` object needs to be just created once in `getConnectionProperties` in this PR.

albinsuresh: It was an attempt to make it explicit at the API level itself that the connection properties are really optional. I could have used null checks. But I prefer the use of `Optional` when something is really optional.

saurabhagas: We could always use an empty `Properties` object -- I understand the trouble with `null`s.

albinsuresh: `Properties` unfortunately doesn't have a static empty instance in it like the `Optional.empty()`. And as I said, explicitly making it optional makes it clear that it's an optional argument and that empty is valid.

saurabhagas: Okay. Can't be too particular about it. Approved.

myronkscott: kinda with Saurabh on this one.  Not a huge fan of the Optional pattern.

PR #948, opened 2018-09-03
[release/5.4] Fix #947: finalize the stack when the client connects during reconnect

### LINKED ISSUE #947: [5.4.2-pre2] server is unable to send responses to clients after reconnect
We found some test failures related to reconnect in PR Terracotta-OSS/terracotta-platform#489 which upgrades terracotta-core version to 5.4.2-pre2

Cloudbees build: https://terracotta-oss.ci.cloudbees.com/job/terracotta-platform-pull-requests/1115/

it looks like the issue got introduced here: https://github.com/Terracotta-OSS/terracotta-core/commit/f4a7cbbd3fe76af53b3e58aeb370f0ea5a36b0c7#diff-2e8b0f445e4261efc3c463cdf9e128e3, channel / client id used in ServerNetworkStackHarness.finalizeStack() is not correct, this is causing issues when server tries to send responses to clients after reconnect.

ramsai1729:
This PR reverts the ServerStackProvider.java changes in https://github.com/Terracotta-OSS/terracotta-core/commit/f4a7cbbd3fe76af53b3e58aeb370f0ea5a36b0c7#diff-2e8b0f445e4261efc3c463cdf9e128e3

### DIFF
--- a/common/src/main/java/com/tc/net/protocol/transport/ServerStackProvider.java
+++ b/common/src/main/java/com/tc/net/protocol/transport/ServerStackProvider.java
@@ -107,7 +107,7 @@ public class ServerStackProvider implements NetworkStackProvider, MessageTranspo
       newStackHarness(clientID, messageTransportFactory.createNewTransport(
           createHandshakeErrorHandler(),
           handshakeMessageFactory,
-          transportListeners)).finalizeStack();
+          transportListeners));
     }
     this.activeProvider = activeProvider;
     this.validateTransport = validate;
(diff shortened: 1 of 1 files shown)


myronkscott:
doing this will re-introduce a connection leak on failover when the reconnect window times out.  Whatever symptoms you are experiencing are probably due to an incomplete patch from master or a merge problem.  We should probably explore the complete fix rather than just reverting.

ramsai1729:
Fixed in #947

PR #949, opened 2018-09-06
Fix #939 - Reconnect needs to rebuild with ConnectionID not ClientID

### LINKED ISSUE #939: ConnectionID leak
If the connection is closed via reconnect timeout, the connection is leaked because ConnectionID equals contract accounts for ServerID as well as ConnectionID

myronkscott:
cherry-picked from master a4768fbca9ae7f08cff4a32553f6b79065b3e0f7

Conflicts:
	common/src/main/java/com/tc/net/protocol/transport/ServerStackProvider.java
	dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java


ramsai1729:
verified failing tc-platform tests reported in  #947, they are passing with this fix.

PR #957, opened 2018-10-01
Fix 954

myronkscott:
Catch all exceptions in message decode


myronkscott:
rolled into separate PR

PR #974, opened 2018-11-19
Relax failover priority configuration mandation for single node servers

albinsuresh:


dso-l2/src/main/java/com/tc/l2/state/ConsistencyManager.java
```diff
@@ -52,6 +52,9 @@ boolean isStateTransition() {
   
   static int parseVoteCount(TcConfig config) {
     Logger consoleLogger = TCLogging.getConsoleLogger();
+    if (config.getServers().getServer().size() == 1) {
+      return -1;
```
ramsai1729: Returning -1 causes unnecessary/wrong warnings like 

2018-11-19 14:47:10,919 WARN - It is recommended to keep the total number of servers and external voters to be an odd number

albinsuresh: Fixed at the source of the issue.

PR #975, opened 2018-11-20
[No Merge] Safe mode startup and diagnostic endpoints to interact with a server …

albinsuresh:
…in safe mode

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
@@ -217,6 +217,7 @@ import java.nio.charset.Charset;
 import java.util.EnumSet;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import org.terracotta.config.TcConfiguration;
 import org.terracotta.entity.BasicServiceConfiguration;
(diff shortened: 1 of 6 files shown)


dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
```diff
@@ -820,11 +823,29 @@ public void handleEvent(ReplicationMessageAck context) throws EventHandlerExcept
...
+    if (configSetupManager.safeModeStartup()) {
+      try {
+        consoleLogger.warn("Waiting in suspended mode");
+        safeMode.await();
+      } catch (InterruptedException e) {
+        throw new RuntimeException("Startup interrupted while waiting in safe mode");
+      }
+    }
+
     startGroupManagers();
```
chaktatha: We may want to put condition here. In some cases, we may want to restart the server

myronkscott:
This approach will likely work however, I'm not sure it is the best approach to halt the main thread on startup like this.  It may be better to hook the consistency manager in a way that prevents the transition to any state until safe mode is exited.  

It seems not quite right to have the group port open but no discovery or election.  This may cause other nodes that are fully open and initiating elections to do undefined things.

albinsuresh:
I had done some testing with the current pause point and even with the group port open it wasn't really affecting other servers that were not paused. The running servers were just not able to discover the servers paused in suspended mode until they were resumed, as expected. The intent was to pause the servers as early as possible with minimal things started. So I guess waiting till the state transition would be a bit too far. I could try and move it to an earlier point before even the group ports are opened.

myronkscott:
> The intent was to pause the servers as early as possible with minimal things started. So I guess waiting till the state transition would be a bit too far. 

So that is a good question.  Why do you want to pause with minimal things started?  It seems it would be appropriate for the server to be fully started but prevented from joining the cluster and stuck in the START state so that the server can be queried or changed, then released to join the cluster.  What is use-cases for pausing early?

PR #976, opened 2018-11-22
Suspended startup for availability mode

albinsuresh:


albinsuresh:
Introduced suspended startup for availability mode and the diagnostic end points to query and interact with suspended servers.

dso-l2/src/main/java/com/tc/l2/state/ConsistencyManagerImpl.java
```diff
@@ -49,6 +51,8 @@
   private final ServerVoterManager voter;
   private final Set<NodeID> activePeers = Collections.synchronizedSet(new HashSet<>());
   private final Set<NodeID> passives = Collections.synchronizedSet(new HashSet<>());
+
+  private volatile boolean forceTransition = false;
```
myronkscott: misnamed.  Suggest something like allowTransition

dso-l2/src/main/java/com/tc/l2/state/SafeModeConsistencyManagerImpl.java
```diff
@@ -0,0 +1,114 @@
+package com.tc.l2.state;
```
myronkscott: Rather than having two separate implementations, can a delegate model be used where common behavior can be pulled into the SafeModeConsistencyManagerImpl (basically the case where a server is transitioning from START->ACTIVE) then delegated to ConsistencyManagerImpl if consistency mode is selected or nothing if availability mode is selected.

dso-l2/src/main/java/com/tc/l2/state/ConsistencyManagerImpl.java
```diff
@@ -263,7 +276,12 @@ public boolean isStuck() {
     @Override
     public Collection<Transition> requestedActions() {
       return getActions();
-    }        
+    }
+
+    @Override
+    public void forceTransition() {
+      ConsistencyManagerImpl.this.forceTransition();
```
myronkscott: If you disable election rerun, you should force a new election here using [startElectionIfNecessary(ServerID.NULL_ID)](https://github.com/Terracotta-OSS/terracotta-core/blob/master/dso-l2/src/main/java/com/tc/l2/state/StateManagerImpl.java#L557)

dso-l2/src/main/java/com/tc/l2/state/ConsistencyMBean.java
```diff
@@ -43,4 +46,9 @@
    * @return the list of transitions requested.
    */
   Collection<Transition> requestedActions();
+
+  /**
+   * Forcibly allow a requested transition action
+   */
+  void forceTransition();
```
myronkscott: This belongs in its own MBean

dso-l2/src/main/java/com/tc/l2/state/SafeModeConsistencyManagerImpl.java
myronkscott: Would be nice too if the server node tracking (GroupListener) could be factored out and shared as well but that can be future cleanup.

dso-l2/src/main/java/com/tc/l2/state/ConsistencyMBean.java
albinsuresh: Kept this here as this would provide a unified way to get a server out of suspended mode on startup. The old override vote mechanism won't work for a server suspended during startup
even in consistency mode. So I thought I'll add it here so that the signalling is uniform for both availability and consistency modes.

dso-l2/src/main/java/com/tc/config/schema/setup/StandardConfigurationSetupManagerFactory.java
```diff
@@ -157,6 +159,10 @@ public static Options createOptions(ConfigMode configMode) {
       l2NameOption.setRequired(false);
       l2NameOption.setArgName("l2-name");
       options.addOption(l2NameOption);
+
+      Option safeModeOption = new Option("s", "safe", false, "to start the server in safe mode");
```
myronkscott: This looks like a new consistency mode.  Availability with safe startup.  It really should be a stripe wide configuration.  It doesn't make much sense to have a single node in this mode and all the others not in this mode.  We can improve this with later, what do you think?

albinsuresh: This option is to be used only on cluster restarts. For the initial start of the cluster this option doesn't make much sense. That's why I kept it as an option. Even on restart it's upto the customer whether to start in this safe mode or the regular high availability mode. So I feel this being an option gives more flexibility.

albinsuresh: After thinking a bit more, having a dedicated consistency mode would be a good idea so that users won't have to worry about providing extra options on startups and even across all servers in all stripes. That's more error prone although it provides better flexibility. We'll definitely discuss and improve this later.

PR #982, opened 2018-12-07
Safe start servers by default in availability mode 

albinsuresh:
with an added option for unsafe startups. Dependent on https://github.com/Terracotta-OSS/galvan/pull/192

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
@@ -551,7 +551,7 @@ public class DistributedObjectServer implements ServerConnectionValidator {
       consoleLogger.warn("It is not recommended to configure external voters when there is an odd number of servers in the stripe");
     }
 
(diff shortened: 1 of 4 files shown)


dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
```diff
@@ -551,7 +551,7 @@ public synchronized void start() throws IOException, LocationNotCreatedException
       consoleLogger.warn("It is not recommended to configure external voters when there is an odd number of servers in the stripe");
     }
 
-    boolean safeStartup = voteCount < 0 ? configSetupManager.safeModeStartup() : true;
+    boolean safeStartup = configSetupManager.allCurrentlyKnownServers().length > 1 && configSetupManager.safeModeStartup();
```
myronkscott: can you use knownPeers > 0 here?

albinsuresh: Done

PR #987, opened 2018-12-17
Fix #984 - Prevent state transitions after a conditional shutdown

### LINKED ISSUE #984: Prevent state transistions after a conditional shutdown request
On receiving conditional shutdown requests on the server, no further state transition should be allowed to safely complete a conditional shutdown. For eg, if a server is to be shutdown using the `TCServer.stopIfPassive()` primitive, the server should not be allowed to transition to an active state while the shutdown request is being processed.

ramsai1729:


dso-l2/src/main/java/com/tc/l2/state/StateManagerImpl.java
```diff
@@ -336,12 +341,13 @@ public synchronized void moveToPassiveStandbyState() {
   }
 
   private synchronized void moveToActiveState() {
+    synchronizedWaitForStart();
```
ramsai1729: @myronkscott any particular reason for not using `synchronizedWaitForStart()` in `moveToActiveState()`, all other state transition methods have this safety check?

ramsai1729:
@myronkscott last three builds were good, could you merge if looks good.

PR #1013, opened 2019-03-06
Replication perf

myronkscott:
some important replication speed-up changes particularly for multi-passive setups.  The most critical is 5a1b188.

### DIFF
--- a/dso-l1/src/main/java/com/tc/object/request/MultiRequestReceiveHandler.java
+++ b/dso-l1/src/main/java/com/tc/object/request/MultiRequestReceiveHandler.java
@@ -24,30 +24,48 @@ import com.tc.async.api.EventHandlerException;
 import com.tc.entity.VoltronEntityMultiResponse;
 import com.tc.object.ClientInstanceID;
 import com.tc.object.tx.TransactionID;
+import com.tc.text.PrettyPrintable;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.LongAdder;
 
 
-public class MultiRequestReceiveHandler extends AbstractEventHandler<VoltronEntityMultiResponse> {
+public class MultiRequestReceiveHandler extends AbstractEventHandler<VoltronEntityMultiResponse> implements PrettyPrintable {
   private final RequestResponseHandler handler;
+  private final LongAdder opCount = new LongAdder();
+  private final LongAdder msgCount = new LongAdder();
 
   public MultiRequestReceiveHandler(RequestResponseHandler handler) {
     this.handler = handler;
   }
 
   @Override
+  public Map<String, ?> getStateMap() {
+    Map<String, Object> map = new LinkedHashMap<>();
+    map.put("opCount", opCount.longValue());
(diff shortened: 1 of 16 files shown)


tc-messaging/src/main/java/com/tc/entity/LinearVoltronEntityMultiResponse.java
```diff
@@ -276,4 +280,11 @@ public synchronized boolean startAdding() {
     }
     return false;
   }
+  
+  private boolean immediateSend = false;
+  
+  @Override
+  public boolean shouldSend() {
```
ramsai1729: should this be synchronized?

myronkscott: no.  the path it is used on is only for a the same thread.  It is only used when the server is in single threaded mod optimization so the thread that is setting it is accessing it.  I'll add a note about that.

dso-l1/src/main/java/com/tc/object/request/MultiRequestReceiveHandler.java
```diff
@@ -24,30 +24,48 @@
...
+import com.tc.text.PrettyPrintable;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.LongAdder;
 
 
-public class MultiRequestReceiveHandler extends AbstractEventHandler<VoltronEntityMultiResponse> {
+public class MultiRequestReceiveHandler extends AbstractEventHandler<VoltronEntityMultiResponse> implements PrettyPrintable {
   private final RequestResponseHandler handler;
+  LongAdder opCount = new LongAdder();
```
ramsai1729: private final?

dso-l2/src/main/java/com/tc/objectserver/entity/ActivePassiveAckWaiter.java
```diff
@@ -60,6 +61,25 @@ public synchronized void waitForReceived() {
...
+        wasSet = true;
+      }
+    }
+    if (!wasSet) {
+      r.run();
+    }
+  }
+  
+  public void run() {
+    if (finalizer != null) {
```
ramsai1729: no visibility issues?

dso-l1/src/main/java/com/tc/object/request/MultiRequestReceiveHandler.java
myronkscott: yes.  will fix

dso-l2/src/main/java/com/tc/objectserver/entity/ActivePassiveAckWaiter.java
myronkscott: No.  Implemented this way to keep the run out of they synchronization block but I will move the execution in this class so that it is more clear when the finalizer is executed.

myronkscott: added a comment about visibility and updated the code

PR #1014, opened 2019-03-14
add gc monitoring

lorban:
Similar to the stages monitoring, this patch will log GC activity and heap occupation every few seconds.

I am not certain we want this in the server, so I have no hard feelings if you don't want to merge this. But this logs the basics to quickly help us figure out if the GC is healthy or not, and if the heap is reasonably sized.


lorban:
Note: I could not test these changes yet because of the `-f` / `-r` merge problem of yesterday. Please don't merge this for now.

PR #1017, opened 2019-03-22
Security auditing for manual server transition

devlg75:


myronkscott:
Transitions include but are not limited to promotions.

devlg75:
> For this feature, while I don't have a good idea of the desired effect, I don't think you are hooked into the right spot.

This piece of code was added to get audit logging for the promote command in cluster tool. Since the promote command results in that allowTransition MBean invocation on the server, that was the only point where you could add this hook.

PR #1025, opened 2019-04-29
Add support for starting the server in diagnostic mode

ramsai1729:


myronkscott:
Why is this a command line option?  Shouldn't we be able to detect internally when this mode is needed?

PR #1032, opened 2019-05-30
Sum perf and fix

myronkscott:
several consolidated perf and bug fixes.


dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
```diff
@@ -415,6 +415,8 @@ public synchronized void start() throws IOException, LocationNotCreatedException
     this.tcProperties = TCPropertiesImpl.getProperties();
     this.l1ReconnectConfig = new L1ReconnectConfigImpl();
     
+    TCByteBufferFactory.setPoolingEnabled(tcProperties.getBoolean(TCPropertiesConsts.BYTEBUFFER_POOLING, true));
```
cschanck: Assume this is not the actual default?

myronkscott: fixed

PR #1036, opened 2019-06-18
Transaction LWM

chrisdennis:


dso-l1/src/main/java/com/tc/object/ClientEntityManagerImpl.java
```diff
@@ -546,25 +551,30 @@ private void internalRelease(EntityID eid, EntityDescriptor entityDescriptor, Ru
...
+        if (inFlight.send()) {
+          //  when encountering a send for anything other than an invoke, wait here before sending anything else
+          //  this is a bit paranoid but it is to prevent too many resends of lifecycle operations.  Just
+          //  make sure those complete before sending any new invokes or lifecycle messages
+          if (inFlight.getMessage().getVoltronType() != VoltronEntityMessage.Type.INVOKE_ACTION) {
+            inFlight.waitForAcks();
           }
         } else {
-          throwClosedExceptionOnMessage(inFlight, "Connection closed before sending message");
+          logger.debug("message not sent.  Make sure resend happens " + inFlight);
```
myronkscott: logic change?  Is this another bug?

dso-l1/src/main/java/com/tc/object/ClientEntityManagerImpl.java
```diff
@@ -574,27 +584,32 @@ private NetworkVoltronEntityMessage createMessageWithoutClientInstance(EntityID
...
-    if (this.channel.getProductID().isReconnectEnabled()) {
-      for (TransactionID pendingID : this.inFlightMessages.keySet()) {
-        if (oldestTransactionPending.compareTo(pendingID) > 0) {
-          // pendingID is earlier than oldestTransactionPending.
-          oldestTransactionPending = pendingID;
+    TransactionID transactionID = transactionSource.create();
+    TransactionID oldestTransactionPending = transactionSource.oldest();
+    //this is expensive -- only assert when testing
+    TransactionID oldestTransactionInFlight;
+    assert (oldestTransactionInFlight = oldestTransactionIn(inFlightMessages.keySet())) == null || (oldestTransactionPending.compareTo(oldestTransactionInFlight) <= 0);
```
myronkscott: are asserts off by default?

dso-l1/src/main/java/com/tc/object/ClientEntityManagerImpl.java
chrisdennis: This is me screwing up.

chrisdennis: Actually no it's not... it's the git diffing engine screwing things up.

dso-l1/src/main/java/com/tc/object/ClientEntityManagerImpl.java
chrisdennis: Yes.... build-system test runners tend to turn them on. I was torn between this and eliminating this check entirely. Oldest transaction lookup is now O(1), rather than O(N) where N is the number of inflight txns.

PR #1043, opened 2019-06-24
 Modifying tests to be compatible with java11

mobasherul:


common/src/test/java/com/tc/net/basic/BasicConnectionTest.java
```diff
@@ -112,6 +113,7 @@ public void testGetConnectTime() throws Exception {
    * Test of getIdleTime method, of class BasicConnection.
    */
   @Test
+  @Ignore
```
mobasherul: This test does not mock things properly and can fail even on master and get caught in infinite loop causing out of memory problems.

myronkscott: Can we fox the test instead?

mobasherul: I will try and fix this in this PR itself

common/src/test/java/com/tc/runtime/MemoryPoolsTest.java
```diff
@@ -33,30 +33,31 @@ public MemoryPoolsTest() {
...
-      assertTrue(collectorCount1 > -1);
-      System.gc();
-      MemoryUsage mu2 = memManager.getOldGenUsage();
-      assertNotNull(mu2);
-      long collectorCount2 = mu2.getCollectionCount();
-      System.err.println("Now the Collector Count  is  " + collectorCount2);
-      assertTrue(collectorCount2 > collectorCount1);
-    }
+    System.out.println("Test");
+//    List<MemoryPoolMXBean> pools = ManagementFactory.getMemoryPoolMXBeans();
```
cljohnso: This is a pretty big block to simply comment out.  Is there an issue with this code?

pom.xml
```diff
@@ -42,7 +42,7 @@
     <skip.deploy>false</skip.deploy>
 
     <terracotta-apis.version>1.6.1</terracotta-apis.version>
-    <terracotta-configuration.version>10.6.2-pre2</terracotta-configuration.version>
+    <terracotta-configuration.version>10.6-SNAPSHOT</terracotta-configuration.version>
```
cljohnso: Should this be merge like this?

common/src/test/java/com/tc/runtime/MemoryPoolsTest.java
mobasherul: I need to debug this

pom.xml
mobasherul: Nope will fix it

dso-l2/src/test/java/com/tc/objectserver/core/impl/ManagementTopologyEventCollectorTest.java
```diff
@@ -48,8 +48,8 @@
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import org.mockito.AdditionalMatchers;
+import static org.mockito.ArgumentMatchers.*;
```
myronkscott: avoid * imports

common/src/test/java/com/tc/runtime/MemoryPoolsTest.java
mobasherul: This test is failing with java11 but passing with java8

mobasherul:
> There are many cases in this commit stream where an import has been added for `ArgumentMatchers` but a qualified reference using `Mockito` is use for a method in `ArgumentMatchers`. Since `Mockito` extends `ArgumentMatchers`, would it not make more sense to stick with `Mockito`?
> 
> There are also cases where casts or explicit type arguments are used when, in context, the cast/type is likely unnecessary.

Mockito is used in places because of how mockito2 handles null values which is different from mockito1 and ArgumentMatcher should be used in place where Matcher is used as it is deprecated now .

common/src/test/java/com/tc/runtime/MemoryPoolsTest.java
```diff
@@ -33,6 +33,7 @@ public MemoryPoolsTest() {
   }
 
   public void testMemoryPools() throws Exception {
+    // This test fails with java11 because of collection count assertion failure .
```
cljohnso: This test is dependent on the GC implementation being used.  Under Java 8, ParallelGC is used -- this implementation responds positively to a `System.gc()` call so the test works.  Under Java 11, G1GC is the default -- this implementation checks to see if it thinks a GC is necessary and, under conditions of the test, doesn't perform a GC -- so the test fails.  The test either needs to apply some heap load so a GC is necessary or needs to check something else when G1GC is being used and test something else (or skip the test entirely).

cljohnso: This test is dependent on the GC implementation being used.  Under Java 8, ParallelGC is used -- this implementation responds positively to a `System.gc()` call so the test works.  Under Java 11, G1GC is the default -- this implementation checks to see if it thinks a GC is necessary and, under conditions of the test, doesn't perform a GC -- so the test fails.  The test either needs to apply some heap load so a GC is necessary or needs to check something else when G1GC is being used and test something else (or skip the test entirely).

cljohnso: On further investigation, it turns out that G1GC under Java 8 _does_ actually perform a GC -- so this is a G1GC change that's related to G1GC in combination with Java 11.

dso-l2/pom.xml
```diff
@@ -131,6 +131,11 @@
       <groupId>commons-cli</groupId>
       <artifactId>commons-cli</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.glassfish.jaxb</groupId>
+      <artifactId>jaxb-runtime</artifactId>
+      <version>2.3.1</version>
+    </dependency>
   </dependencies>
```
myronkscott: we cannot add this.  Can the code requiring this be removed or modified?

dso-l2/src/test/java/com/tc/objectserver/core/impl/ManagementTopologyEventCollectorTest.java
```diff
@@ -163,7 +165,7 @@ public void testCreatePromoteDestroy() throws Exception {
     long consumerID = 1;
     // Note that we start in passive state.
     boolean isActive = false;
-
```
myronkscott: These whitespace changes clutter the PR.  Is it possible to move them to a separate commit or PR?

dso-l2/pom.xml
mobasherul: All the tests integration-tests within galvan-support module starts failing without this dependency
 java.lang.NoClassDefFoundError: javax/xml/bind/JAXBException
[ClientLeakIT][testServer0] 	at com.tc.config.DefaultConfigurationProvider.getTcConfiguration(DefaultConfigurationProvider.java:140)
[ClientLeakIT][testServer0] 	at com.tc.config.DefaultConfigurationProvider.initialize(DefaultConfigurationProvider.java:106)
[ClientLeakIT][testServer0] 	at com.tc.server.TCServerMain.main(TCServerMain.java:66)
[ClientLeakIT][testServer0] Caused by: java.lang.ClassNotFoundException: javax.xml.bind.JAXBException
[ClientLeakIT][testServer0] 	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:583)
[ClientLeakIT][testServer0] 	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
[ClientLeakIT][testServer0] 	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
[ClientLeakIT][testServer0] 	... 3 more

dso-l2/src/test/java/com/tc/objectserver/core/impl/ManagementTopologyEventCollectorTest.java
mobasherul: I will remove this extra white spaces in this PR itself.

dso-l2/pom.xml
mobasherul: There are some tests failing in galvan module because of java.lang.NoClassDefFoundError: javax/xml/bind/JAXBException when running with java11 at the time of startup of server.

myronkscott: I think that change belongs in tc-configuration, not in dso-l2.  There should be no jab requirements tc-core.

mobasherul: Removed this dependency we don't need it

common/src/test/java/com/tc/runtime/MemoryPoolsTest.java
mobasherul: I have added comment here related to GC .

common/src/test/java/com/tc/runtime/MemoryPoolsTest.java
```diff
@@ -33,6 +33,8 @@ public MemoryPoolsTest() {
...
+    // It depends on the GC algortihm whether it will trigger GC or not on calling System.gc().
```
cljohnso: The `assertTrue(collectorCount2 > collectorCount1);` at line 61 needs to be commented out and the comment at this line moved there to explain the assertion's removal.  The assertion is what's bogus about this test -- actual GC is not guaranteed when `System.gc()` is called and the assertion is based on the fallacy that GC will run when `System.gc()` is called.

PR #1050, opened 2019-07-01
Group error handling

myronkscott:
Server message completion refactoring

Election deadlock fix


myronkscott:
Self merging in the interest of time

PR #1056, opened 2019-07-23
async invocation API implementation

lorban:
Please pay attention to the new tests in ClientEntityManagerTest.

### DIFF
--- a/dso-l1/src/main/java/com/tc/object/ClientEntityManagerImpl.java
+++ b/dso-l1/src/main/java/com/tc/object/ClientEntityManagerImpl.java
@@ -71,6 +71,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.LongAdder;
@@ -120,12 +121,12 @@ public class ClientEntityManagerImpl implements ClientEntityManager {
     }
   } 
   
-  private synchronized boolean enqueueMessage(InFlightMessage msg) {
+  private synchronized boolean enqueueMessage(InFlightMessage msg) throws RejectedExecutionException {
     if (!this.stateManager.isRunning()) {
       return false;
     }
     if (!requestTickets.messagePendingSlotAvailable()) {
-      throw new IllegalStateException("Output queue is full");
+      throw new RejectedExecutionException("Output queue is full");
     }
 
     inFlightMessages.put(msg.getTransactionID(), msg);
@@ -293,13 +294,13 @@ public class ClientEntityManagerImpl implements ClientEntityManager {
   }
 
   @Override
-  public void asyncInvokeAction(EntityID eid, EntityDescriptor entityDescriptor, Set<VoltronEntityMessage.Acks> requestedAcks, InFlightMonitor monitor, boolean requiresReplication, byte[] payload) {
+  public void asyncInvokeAction(EntityID eid, EntityDescriptor entityDescriptor, Set<VoltronEntityMessage.Acks> requestedAcks, InFlightMonitor monitor, boolean requiresReplication, byte[] payload) throws RejectedExecutionException {
     asyncQueueInFlightMessage(eid,
         ()->createMessageWithDescriptor(eid, entityDescriptor, requiresReplication, payload, VoltronEntityMessage.Type.INVOKE_ACTION, requestedAcks),
         requestedAcks, monitor);
   }
 
-  private void asyncQueueInFlightMessage(EntityID eid, Supplier<NetworkVoltronEntityMessage> message, Set<VoltronEntityMessage.Acks> requestedAcks, InFlightMonitor monitor) {
+  private void asyncQueueInFlightMessage(EntityID eid, Supplier<NetworkVoltronEntityMessage> message, Set<VoltronEntityMessage.Acks> requestedAcks, InFlightMonitor monitor) throws RejectedExecutionException {
     InFlightMessage inFlight = new InFlightMessage(eid, message, requestedAcks, monitor, false);
     try {
       msgCount.increment();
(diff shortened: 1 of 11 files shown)


dso-l1/src/main/java/com/tc/object/InFlightMessage.java
```diff
@@ -336,23 +336,31 @@ public synchronized void setResult(byte[] value, EntityException error) {
       this.pendingAcks.clear();
       this.exception = error;
       this.getCanComplete = true;
+      handleException(error);// TODO: why isn't handleException called somewhere else?
```
myronkscott: This will not work for the synchronous mode.  In synchronous mode, the monitor does not receive the result as that is returned by delivery of the future.

dso-l1/src/main/java/com/tc/object/ClientEntityManagerImpl.java
```diff
@@ -120,6 +120,19 @@ public boolean checkBusy() {
     }
   } 
   
+  private synchronized boolean enqueueMessage(InFlightMessage msg) {
+    if (!this.stateManager.isRunning()) {
+      return false;
+    }
+    if (!requestTickets.messagePendingSlotAvailable()) {
+      throw new IllegalStateException("Output queue is full");
```
myronkscott: This seems risky to me, there is no better way to handle it?

dso-l1/src/main/java/com/tc/object/InFlightMonitor.java
```diff
@@ -39,6 +39,14 @@ public InFlightMonitor(MessageCodec<?, R> codec, InvokeMonitor<R> monitor, Execu
...
   }
 
+  public void exception(EntityException ee) {
+    if (executor != null) {
+      executor.execute(()->monitor.exception(ee));
+    } else {
+      monitor.exception(ee);
+    }
+  }
+
```
myronkscott: Probably need to add something for results as well.

dso-l1/src/main/java/com/tc/object/ClientEntityManagerImpl.java
lorban: `IllegalStateException` will eventually be caught in `EntityClientEndpointImpl.invoke()` and call `InvocationCallback.failure()`.

This seemed like the only thing to do in async mode: report to the callback that the output queue is full. Do you have a better proposition in mind?

chrisdennis: I think it might be better to throw this directly at the caller rather than report it. It's really a distinct *kind* of failure, right?

lorban: Is it really that distinct? After all, you pushed a message through the API and it failed to be processed. How is that any different than if the client could failed to deliver the message to the server?

chrisdennis: I guess my main justification here is that it's synchronously generated which might have implications for the callback logic (although that would be grounds for throwing all synchronously generated exceptions).

lorban: Would you prefer an API where you have to handle exceptions both with a try-catch around `AsyncInvocationBuilder.invoke()` _and_ in `InvocationCallback.failure()` or just in the latter? My personal preference goes for the 2nd option.

If we can't agree, we should involve more team members.

dso-l1/src/main/java/com/tc/object/InFlightMessage.java
lorban: I think I see what you mean: `InFlightMessage` should discriminate between sync and async mode and only call the two `handle*` methods in the async case, and `notifyAll` in the sync case. Let me try this.

dso-l1/src/main/java/com/tc/object/ClientEntityManagerImpl.java
myronkscott: Maybe throw RejectedExecutionException which gets thrown back to the caller of ayncInvoke()?  Seems better to let the caller know that nothing ever got started rather than having the callback handle it.  My concern is that the linear sequence gets all messed up if you don't handle this at the caller of asyncInvoke.  It might not matter in most cases but I could see situations where it matters.

lorban: Okay, I trust your instinct. I'll make the code throw `RejectedExecutionException` and declare that as `throws` all along the stack.

dso-l1/src/main/java/com/tc/object/InFlightMessage.java
lorban: I've handled this case by discriminating between sync and async when creating the `InFlightMessage`. Hacky, but didn't require much engineering.

dso-l1/src/main/java/com/tc/object/InFlightMonitor.java
lorban: Same as above: `InFlightMonitor` now discriminates between sync and async.

lorban:
Assuming we're happy with the new API, there is a ton of things that could be cleaned up and refactored, like reworking the sync impl on top of the async one. I would prefer that we move forward using this new API first.

chrisdennis:
I think this makes sense with one caveat.  We know that adoption of this API is unlikely to happen in the 10.5.0 release... are we comfortable that there are no adverse effects that could be seen by non-users?

dso-l1/src/main/java/com/tc/object/ExceptionUtils.java
```diff
@@ -46,9 +46,8 @@ public static EntityException addLocalStackTraceToEntityException(EntityID eid,
     } else if(e instanceof EntityBusyException) {
       wrappedException = new EntityBusyException(entityType, enityName, e);
     } else {
-//  just return the remote exception with remote stack, don't want to hide the type of the 
-//  exception in these cases.
-      return e;
+      // TODO we can't just return the exception as this would hide some of the stack, it must be wrapped in a new EntityException type
```
myronkscott: This change worries me a bit.  I believe some code relies on the type exposed here.

lorban: This may break some code, and more probably some tests. I nevertheless believe this is a good change, but outside the scope of this PR so I'm going to revert it.

PR #1060, opened 2019-09-17
Fix permanent entity creation during rolling upgrades

ramsai1729:
Looks like the server creates permanent entities only if its a fresh startup, if we add new permanent entities in the new version, they never get created unless we wipe the existing data

### DIFF
--- a/dso-l2/src/main/java/com/tc/objectserver/api/EntityManager.java
+++ b/dso-l2/src/main/java/com/tc/objectserver/api/EntityManager.java
@@ -29,6 +29,7 @@ import com.tc.text.PrettyPrintable;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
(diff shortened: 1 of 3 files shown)


myronkscott:
I think this change requires further discussion.

myronkscott:
This will not work way around the problem will not work.  Permanent entities are created at a very specific time in the lifecycle.  These entities are created only at initial fresh server creation before clients connect and before any passives connect.  This guarantees that each passive that subsequently connects has an exact copy of the entity and entity set.  If this change is allowed, 3 node stripes will be broken.  For example, if two passives are started with new permanent entity jars, once the active is killed, a new permanent entity will be created on the next active.  This new entity will not be present on the current passive that connects to the new active.  Any changes requiring replication will result in entity not found.  Any changes not requiring replication will not be on the next passive if it goes active and re-creates the permanent entity.

PR #1064, opened 2019-09-20
Batch script cleanup

chrisdennis:


myronkscott:
Does this need to be targeted at the 5.6 release line as well?  Currently it is only merged to 5.7 (trunk)

PR #1071, opened 2019-11-13
Added ability to set client-reconnect-window in TerracottaServer MBean

saurabhagas:


dso-l2/src/main/java/com/tc/server/TCServer.java
```diff
@@ -76,6 +76,8 @@
   
   int getReconnectWindowTimeout();
 
+  void updateReconnectWindowTimeout(int value);
```
myronkscott: I'd prefer we stick with the set<property> convention across the PR.

saurabhagas: I had done the same earlier, but I discovered that methods starting with "get" and "set" are considered getters and setters for attributes, and not considered operations.

myronkscott: are we not changing an attribute here?

mathieucarbou: Saurabh: whether it is exposed as an attribute or operation in JMX, it does not change anything for us: we can invoke both an operation or change an attribute in JMX.
But of it was exposed as an operation before, because of the missing setter, then perhaps it is better to still expose them as operations if we want to keep backward compatibily

saurabhagas: @myronkscott, I was referring to the MBean's conventions of what are considered attributes and operations. A method called `getX` and not accepting an argument is considered a getter for attribute X. Such methods can be called using `MBeanServer::getAttribute` instead of `MBeanServer::invoke`. I wanted this method to be treated as an operation so that it can be called using `MBeanServer::invoke`, but like @mathieucarbou pointed out, it doesn't matter, so I've used the `set` prefix.

@mathieucarbou, a missing setter doesn't convert an attribute into an operation - it just makes the attribute read-only. So, before this PR `getReconnectWindowTimeout` exposed a read-only attribute called `ReconnectWindowTimeout`. 

As to your second point, I'm in for introducing a new MBean which is the right abstraction for making dynamic changes. I'll be adding the capability to add support for updating hostname etc. very soon, and will take the best available approach. The changes in this PR are fine for now IMO (subject to refactoring in near future).

mathieucarbou: Fine for me, I'll let @myronkscott or @ramsai1729 merge :-)

dso-l2/src/main/java/com/tc/objectserver/handshakemanager/ServerClientHandshakeManager.java
```diff
@@ -286,7 +288,7 @@ synchronized int getUnconnectedClientsSize() {
     private ReconnectTimerTask(ServerClientHandshakeManager handshakeManager, Timer timer) {
       this.handshakeManager = handshakeManager;
       this.timer = timer;
-      this.timeToWait = handshakeManager.reconnectTimeout;
+      this.timeToWait = handshakeManager.reconnectTimeoutSupplier.get();
```
ramsai1729: `ReconnectTimerTask` and `ServerClientHandshakeManager` might not have the same timeout value, I think its better to create the ` ReconnectTimerTask` task  in `startReconnectWindow` passing the timeout value instead of creating in the constructor

saurabhagas: Done

dso-l2/src/main/java/com/tc/objectserver/handshakemanager/ServerClientHandshakeManager.java
```diff
@@ -307,8 +304,7 @@ public void run() {
 
         if (this.timeToWait < RECONNECT_WARN_INTERVAL) {
           cancel();
-          final ReconnectTimerTask task = new ReconnectTimerTask(this.handshakeManager, this.timer);
-          task.setTimeToWait(this.timeToWait);
+          final ReconnectTimerTask task = new ReconnectTimerTask(this.handshakeManager, this.timer, timeToWait);
```
ramsai1729: maybe `this.timeToWait` for consistency

saurabhagas: Done

PR #1102, opened 2020-01-21
Fix build on mac

ramsai1729:
Build is failing on my mac with below error 

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:3.2.0:single (create-distribution) on project default-configuration: Execution create-distribution of goal org.apache.maven.plugins:maven-assembly-plugin:3.2.0:single failed: group id '1978728554' is too big ( > 2097151 ). Use STAR or POSIX extensions to overcome this limit -> [Help 1]
[ERROR] 

Build is fine with the changes in this PR

### DIFF
--- a/default-configuration/pom.xml
+++ b/default-configuration/pom.xml
@@ -115,6 +115,7 @@
                     <goal>single</goal>
                   </goals>
                   <configuration>
+                    <tarLongFileMode>posix</tarLongFileMode>
                     <descriptorRefs>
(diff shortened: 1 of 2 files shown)


ramsai1729:
This issue still happens on my machine and looks like suggested solution is to use `posix` mode. terracotta-platform uses `posix` mode as well.

@myronkscott can you share the issues you found with `posix` mode?

ramsai1729:
Fixed in #1156

PR #1108, opened 2020-02-12
Client side configuration cleanup

ramsai1729:
* Removes unnecessary classes on client side
* Iterable<InetSocketAddress> is given in `ConnectionFactory.connect` passed down to `ClientConnectionEstablisher` - this helps in propagating topology changes


common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
```diff
@@ -56,7 +53,7 @@
   private static final long                 MIN_RETRY_INTERVAL    = 1000;
   public static final String                RECONNECT_THREAD_NAME = "ConnectionEstablisher";
 
-  private final LinkedHashSet<ConnectionInfo>   connAddressProvider;
+  private volatile Iterable<InetSocketAddress> serverAddresses;
```
myronkscott: why is this volatile?  it is set in a synchronized block.

common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
```diff
@@ -118,63 +114,56 @@ boolean getAllowReconnects() {
...
+        target = nextServerAddress(serverAddressIterator);
       }
       try {
-        return cmt.open(info);
+        return cmt.open(target);
       } catch (TransportRedirect redirect) {
-        ConnectionInfo add = new ConnectionInfo(redirect.getHostname(), redirect.getPort());
-        reporter.onError(info, redirect);
-        info = add;
-        if (this.connAddressProvider.add(add)) {
```
myronkscott: removing this causes redirects to be transient.  We want redirects to be permanent additions to the list of known servers.

common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
ramsai1729: but not all reads under the synchronized block.

common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
ramsai1729: this is not a problem If the user has specified all the servers in the stripe, right?

common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
myronkscott: toString reads don't matter.  I don't volatile is warranted here.

common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
myronkscott: true but that isn't going to happen all the time.

myronkscott: plus that ability was there before.  we should preserve it.

ramsai1729: Made some changes to preserve redirects, please have a look

common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
ramsai1729: what about reconnect(...)?

ramsai1729:
@myronkscott PR rebased

PR #1113, opened 2020-03-05
Adapt to latest Galvan changes for dynamic-config

saurabhagas:


default-configuration/pom.xml
```diff
@@ -110,6 +110,7 @@
                     <goal>single</goal>
                   </goals>
                   <configuration>
+                    <tarLongFileMode>posix</tarLongFileMode>
```
myronkscott: posix breaks on long file names. how is GNU failing for you?

galvan-support/src/main/java/org/terracotta/testing/config/TcConfigConvertor.java
```diff
@@ -0,0 +1,75 @@
+/*
```
myronkscott: Can this be moved to tc-platform?  It doesn't seem to be used here.

terracotta-kit/pom.xml
```diff
@@ -135,7 +135,7 @@
                     <finalName>terracotta-${project.version}</finalName>
                     <appendAssemblyId>false</appendAssemblyId>
                     <attach>true</attach>
-                    <tarLongFileMode>gnu</tarLongFileMode>
```
myronkscott: This change breaks packaging on my local machine

pom.xml
```diff
@@ -43,7 +43,7 @@
 
     <terracotta-apis.version>1.7.0-pre1</terracotta-apis.version>
     <terracotta-configuration.version>10.7.0-pre2</terracotta-configuration.version>
-    <galvan.version>1.4.4-pre7</galvan.version>
+    <galvan.version>1.4-SNAPSHOT</galvan.version>
```
myronkscott: 1.4.4-pre8

terracotta-kit/pom.xml
saurabhagas: Got checked-in by mistake. Will remove.

default-configuration/pom.xml
saurabhagas: Got checked-in by mistake. Will remove.

pom.xml
saurabhagas: Cool, thanks

galvan-support/src/main/java/org/terracotta/testing/config/TcConfigConvertor.java
saurabhagas: It's being used in `BasicExternalCluster`

myronkscott: can you move this closer to tc-platform where the tool actually lives?

galvan-support/src/main/java/org/terracotta/testing/rules/BasicExternalCluster.java
```diff
@@ -68,30 +84,31 @@
...
   private boolean isSafe;
 
-  BasicExternalCluster(File clusterDirectory, int stripeSize, List<File> serverJars, String namespaceFragment,
-                       String serviceFragment, int clientReconnectWindowTime, int failoverPriorityVoterCount,
-                       Properties tcProperties, Properties systemProperties1, String logConfigExt) {
-    if (clusterDirectory.exists()) {
-      if (clusterDirectory.isFile()) {
+  BasicExternalCluster(Path clusterDirectory, int stripeSize, Set<Path> serverJars, String namespaceFragment,
+                       String serviceFragment, int clientReconnectWindow, int voterCount, Properties tcProperties,
+                       Properties systemProperties, String logConfigExt, boolean convertConfig) {
```
myronkscott: Pass in the startup builder here.  The legacy one lives here with galvan-support code.  The new dynamic startup builder lives with dynamic config in tc-platform

galvan-support/src/main/java/org/terracotta/testing/config/TcConfigConvertor.java
saurabhagas: Done

galvan-support/src/main/java/org/terracotta/testing/rules/BasicExternalCluster.java
saurabhagas: Done. I used functions so that this core truly remains agnostic of the builders and convertors, and purely operates on data. Not proud of the ari-ness of the functions though. I can define interfaces in tc-core instead of defining these new functions. Let me know what you think.

myronkscott: How about some kind of builder interface.

String[] startupArgs = new StartupBuilderImpl().kitLocation(Path).workingDirectory(Path).consistentStartup(boolean).build()

That way you can have some sane defaults in tc-core and tc-platform.  You can define the interface in galvan support.

saurabhagas: Had to rethink my approach for this. Good idea - cleaned up a lot of the mess I had created. Done!

PR #1116, opened 2020-03-10
consistency on startup moves to startup builder

myronkscott:


galvan-support/src/main/java/org/terracotta/testing/config/StartupBuilder.java
```diff
@@ -22,8 +22,9 @@
 
 public interface StartupBuilder {
   StartupBuilder tcConfig(Path tcConfig);
-  StartupBuilder serverInstallationDir(Path serverInstallationDir);
-  StartupBuilder testParentDir(Path testParentDir);
+  StartupBuilder serverKitDir(Path serverInstallationDir);
```
saurabhagas: Variable name can match the API to avoid naming confusion. Also, `kitDir()` should suffice for the API name as the kit directory is the same for the whole test suite.

galvan-support/src/main/java/org/terracotta/testing/config/StartupBuilder.java
```diff
@@ -22,8 +22,9 @@
...
+  StartupBuilder serverLoggingExtension(String logging);
```
saurabhagas: `logging` could be renamed to `loggingExtension` or something.

Also, the name `loggingExtension` suffices for the API as it's all server side. The reason we had "server" prefix for working directory is that there's a unique working directory per server.

galvan-support/src/main/java/org/terracotta/testing/rules/BasicExternalCluster.java
```diff
@@ -180,30 +182,29 @@ private void internalStart() throws Throwable {
...
 
-    StripeConfiguration stripeConfig = new StripeConfiguration(kitDir.toAbsolutePath(), stripeInstallationDir,
-        serverDebugPorts, serverPorts, serverGroupPorts, serverNames, stripeName, serverJars, DEFAULT_SERVER_HEAP_MB,
+    Path kitLocation = installKit(stripeVerboseManager, kitDir, serverJars, stripeInstallationDir);
+
+    StripeConfiguration stripeConfig = new StripeConfiguration(serverDebugPorts, serverPorts, serverGroupPorts, serverNames, stripeName, DEFAULT_SERVER_HEAP_MB,
         logConfigExt, systemProperties);
     StripeInstaller stripeInstaller = new StripeInstaller(interlock, stateManager, stripeVerboseManager, stripeConfig);
     // Configure and install each server in the stripe.
     for (int i = 0; i < stripeConfig.getServerNames().size(); ++i) {
```
saurabhagas: Unrelated to this PR, but `stripeConfig.getServerNames().size()` can be simplified to just say `stripeSize`

PR #1117, opened 2020-03-11
Getters in DefaultStartupCommandProvider + version bumps + test dir name fix

saurabhagas:


galvan-support/src/main/java/org/terracotta/testing/config/StartupBuilder.java
```diff
@@ -1,32 +0,0 @@
-/*
```
myronkscott: Keep the interface. It is less code change

galvan-support/src/main/java/org/terracotta/testing/config/DefaultStartupBuilder.java
```diff
@@ -1,121 +0,0 @@
-/*
```
myronkscott: Rename to abstract and all instance accessors

galvan-support/src/main/java/org/terracotta/testing/config/StartupBuilder.java
saurabhagas: I see two possibilities:
1. we have an interface and tc-core and tc-platform implement the methods with no commonality between them. This is ideal when there's no common code.
2. We have the common code in the abstract class, and only the different part in the subclasses. Everything apart from the `build()` method is the same, so I went for an abstract class.

myronkscott: It’s always nice to program against interfaces event if you have an abstract for convenience and code reuse.  List -> AbstractList is a good example in the jerk collections library.

saurabhagas: Kept the interface as is, and added getters in `DefaultStartupCommandBuilder` so that subclasses can use them.

PR #1120, opened 2020-03-16
Config provider api

myronkscott:


dso-l2/src/main/resources/logback.xml
```diff
@@ -1,8 +1,8 @@
 <configuration>
 
   <appender name="TC_BASE" class="com.tc.l2.logging.BufferingAppender">
-    <target>System.out</target>
-    <console>true</console>
+    <!--target>System.out</target-->
```
saurabhagas: We could remove these lines instead of commenting them out. If it makes sense, you could add a line mentioning why this class shouldn't log to console.

myronkscott: how about I comment on why they are commented out?  I don't want to remove them just yet.  I'm a little worried some downstream consumer is looking for certain things in the console output for tests.

configuration-provider/src/main/java/org/terracotta/configuration/Configuration.java
```diff
@@ -24,7 +24,7 @@
 import java.util.Properties;
 
 public interface Configuration {
-  ServerConfiguration getDefaultServerConfiguration(String serverName)throws ConfigurationException;
+  ServerConfiguration getServerConfiguration() throws ConfigurationException;
```
mathieucarbou: makes sense :-)

dso-l2/src/main/resources/logback.xml
saurabhagas: Sure, up to you.

dso-l2/src/main/java/com/tc/server/CommandLineParser.java
```diff
@@ -37,13 +37,11 @@
 
 import static com.tc.server.CommandLineParser.Opt.CONSISTENT_STARTUP;
 import static com.tc.server.CommandLineParser.Opt.HELP;
-import static com.tc.server.CommandLineParser.Opt.SERVER_NAME;
 import static com.tc.server.CommandLineParser.Opt.UPGRADE_MODE;
 
 class CommandLineParser {
 
   enum Opt {
-    SERVER_NAME("n", "name"),
```
mathieucarbou: makes sense :-)

PR #1131, opened 2020-04-09
Improve behavior of PortChooser

cljohnso:
This commit updates PortChooser to:
* Warn when the candidate port pool is less that 1024 ports; a small
  candidate port pool can result in test timeouts.
* Correct logic in isPortUsed resulting in a failure to fully check
  port availability.
* Introduce a connection timeout for the "client connection" test to
  reduce the time taken to test a port.

### DIFF
--- a/common/src/main/java/com/tc/util/PortChooser.java
+++ b/common/src/main/java/com/tc/util/PortChooser.java
@@ -21,26 +21,56 @@ package com.tc.util;
 import com.tc.net.EphemeralPorts;
 import com.tc.net.EphemeralPorts.Range;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.IOException;
 import java.net.BindException;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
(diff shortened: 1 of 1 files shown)


cljohnso:
I have an instrumented version of `PortChooser` I used for local profiling.  With the changes in this commit, it went from just over 1s per allocation request (no matter the number of ports because of the testing error) to ~50ms per port -- asking for 8 ports takes ~0.4s.

myronkscott:
With the support module now in tc-platform. I think that leaves this change orphaned.

saurabhagas:
> With the support module now in tc-platform. I think that leaves this change orphaned.

Yes. The `galvan-platform-support` in tc-platform is using the new port chooser.

cljohnso:
> > With the support module now in tc-platform. I think that leaves this change orphaned.
> 
> Yes. The `galvan-platform-support` in tc-platform is using the new port chooser.

The port chooser that was in terracotta-platform has a gap that this one fills.  I'll look at merging the capabilities of both into a version resident in terracotta-utilities.

PR #1132, opened 2020-04-14
new parent, surefire/failsafe->forge

akomakom:
latest terracotta-parent, using maven-forge-plugin for full test jvm from toolchains support.  All references for surefire and failsafe are removed (forge plugin is a superset of both).


akomakom:
@myronkscott I think that this is good to go.

PR #1136, opened 2020-04-15
Use latest topology during re-registration

ramsai1729:
Currently host-ports passed on command-line used for re-registration, this might not work as topology might have changed and those host-ports are not part of the stripe.


myronkscott:
please add a test when time permits

PR #1137, opened 2020-04-15
Logging PID for voter process

mobasherul:
Logging PID for voter process


myronkscott:
please rebase

voter/src/main/java/org/terracotta/voter/ActiveVoter.java
```diff
@@ -96,6 +96,7 @@ public ActiveVoter start() {
 
   private Thread voterThread(CompletableFuture<VoterStatus> voterStatus, Optional<Properties> connectionProps, String... hostPorts) {
     return new Thread(() -> {
+      writePID();
```
ramsai1729, line 99: I think its better to dump PID in TCVoterMain as it is useful for script invocations mostly.

PR #1140, opened 2020-04-16
Enforce licenses

akomakom:
core never enforced header licenses due to a pom error.  
Using new parent that enforces licenses. 
Due to a change in license plugin usage, a blank line is removed from all files.


akomakom:
@myronkscott I have these options:
1. Disable spotbugs for now and deal with fixing violations in a separate PR
2. Fix them in this PR

The downside of option 2 is that the diff is already overwhelming.  But it's fine with me.

myronkscott:
Disable spot bugs for this repo. Too much legacy to deal with. I don’t think it is worth it.

PR #1146, opened 2020-05-04
small stability items

myronkscott:
Voter need to be tolerant of errors in registration 
BufferManagerFactorySupplier should always supply at least a ClearTextBufferManager

### DIFF
--- a/voter/src/test/java/org/terracotta/voter/ActiveVoterTest.java
+++ b/voter/src/test/java/org/terracotta/voter/ActiveVoterTest.java
@@ -23,6 +23,7 @@ import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
+import org.mockito.Mockito;
 import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.clearInvocations;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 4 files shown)


voter/src/test/java/org/terracotta/voter/ActiveVoterTest.java
```diff
@@ -255,7 +257,7 @@ public void testReregistrationWhenAllStaticHostPortsNotAvailable() throws Timeou
     promote(passiveManager);
 
     waitForLogMessage("Vote owner state: ACTIVE-COORDINATOR");
-
+    status.get().awaitRegistrationWithAll();
```
ramsai1729, line 260: why do we need this?

voter/src/test/java/org/terracotta/voter/ActiveVoterTest.java
```diff
@@ -278,7 +280,7 @@ private void disconnectManagers(Collection<ClientVoterManager> managers) throws
         Thread.sleep(500);
         return false;
       });
-      when(manager.registerVoter(VOTER_ID)).thenReturn(-1L);
+      Mockito.doReturn(1L).when(manager).registerVoter(VOTER_ID);
```
ramsai1729, line 283: negative value makes the disconnection intention clear.

myronkscott: this was a transcription error.  doReturn is used because it works better with spy

voter/src/test/java/org/terracotta/voter/ActiveVoterTest.java
myronkscott: I was mis-interpreting a test failure.  Removed

PR #1147, opened 2020-05-04
Remove no-longer-used tc-config.xml and add new properties file 

jhouserizer:
Remove no-longer-used tc-config.xml and add new example properties file


myronkscott:
The base kit production going forward is here now

https://github.com/Terracotta-OSS/terracotta-platform/tree/master/kit

This is legacy and scheduled for removal at some point

jhouserizer:
doh!

jhouserizer:
obsolete

PR #1150, opened 2020-05-26
Fix 1149

myronkscott:
Fix connection deadlock and remove unused OOO network protocol.


chrisdennis:
Since we're removing TC properties there are probably some associated `TCPropertiesConsts` and `tc.properties` changes.

PR #1161, opened 2020-07-27
Fix the signature of passive mbean changes

myronkscott:
Update the signature of passive addition and removal to include group port.


ramsai1729:
@myronkscott DynamicPassiveRemovalIT is failing, looks like we need to JMX invocation.

PR #1162, opened 2020-07-28
Use new Server::audit API for auditing

saurabhagas:


dso-l2/src/main/java/com/tc/objectserver/impl/DiagnosticsHandler.java
```diff
@@ -95,11 +95,7 @@ private void processMessage(TCMessage message) {
           result = initialState != null ? initialState.getName().getBytes(set) : "".getBytes(set);
           break;
         case "getClusterState":
-          if (GuardianContext.validate(Guardian.Op.SERVER_DUMP, "getClusterState")) {
-            result = server.getClusterState(set, null);
-          } else {
-            result = "NOT PERMITTED".getBytes(set);
-          }
+          result = server.getClusterState(set, null);
```
saurabhagas, line 98: This operation wasn't a dump operation - it was "get me the dumps" operation, something we don't need to guard (and thus audit)

dso-l2/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java
```diff
@@ -37,7 +37,6 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terracotta.entity.PlatformConfiguration;
-import org.terracotta.entity.ServiceConfiguration;
```
saurabhagas, line 40: Removed unused imports while I was here

dso-l2/src/main/java/com/tc/server/TCServerImpl.java
```diff
@@ -137,6 +137,7 @@ public void stopIfActive(StopAction...restartMode) throws PlatformStopException
 
   @Override
   public void stop(StopAction...restartMode) {
+    GuardianContext.validate(Guardian.Op.STATE_CHANGE, "Stop invoked");
```
saurabhagas, line 140: Put the `validate` call in the right place, so that no matter where it's called from, we guard the operation

server-spi/src/main/java/com/tc/spi/Guardian.java
```diff
@@ -20,25 +20,26 @@
...
     ENTITY_DESTROY,
-    // false return on these operations will result in a refusal of the network connection
-    CONNECT_SERVER,
-    CONNECT_CLIENT,
-    
-    SERVER_DUMP,
-    SERVER_EXIT,
+
+    // Change categories. Cannot be validated - false return on validate() will not be honored
+    TOPOLOGY_CHANGE,
```
saurabhagas, line 36: Introduced meaningful categories. `SERVER_DUMP` and `SERVER_EXIT` were too specific, and `GENERIC_OP` is too vague. All the new categories warrant security auditing, and `GENERIC_OP` can be reserved for everything else.

saurabhagas:
Verified these changes in the downstream branches.

Also, please include this change in the next tc-core release, so that I don't need to start a train just for this change.

dso-l2/src/main/java/com/tc/l2/state/ConsistencyMBeanImpl.java
```diff
@@ -49,6 +51,7 @@ public boolean isStuck() {
 
   @Override
   public void allowRequestedTransition() {
+    GuardianContext.validate(Guardian.Op.STATE_CHANGE, "Promote invoked");
     consistencyManager.allowLastTransition();
```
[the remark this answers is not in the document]
saurabhagas: I understand - I was trying to have IDs that contain exactly what we want to audit, so that we don't have to speical-case a bunch of things. This isn't right, I know. Could we add a new `message` parameter in the `Guardian::validate` method so that I could achieve this?

In addition to adding a new parameter to `Guardian::validate`, I could change "Promote invoked" to something like "Request for server state transition to X state received". How does that sound? Also, what's the easiest way to get the new state X?

dso-l2/src/main/java/com/tc/objectserver/impl/DiagnosticsHandler.java
saurabhagas: The intention of auditing cluster tool dump was to avoid abuse of server logging (somebody could fill up the server logs with endless backup data). The analogous action from TMC is to fetch the server dumps in the browser and view/save them. This operation is a non-privileged operation in TMC, and was never being audited. I, thus, removed this guard, especially that `SERVER_DUMP` operation is gone.

saurabhagas:
@myronkscott updated the PR after your auditing changes. PR is a lot simplified now. Please include this change in the next core release so that downstream repositories can use this change.

PR #1167, opened 2020-08-11
put messages on queue even when paused to handle failovers

myronkscott:


saurabhagas:
Failing `FailoverIT` in platform passes with this change. Merging the PR and triggering a release.

PR #1175, opened 2020-09-02
Add a kill switch 

myronkscott:
An uncaught exception on a terracotta thread should result in connection closure.


ramsai1729 (approved):
Looks like we can remove shutdown from below code, can be done later.
```
Reference<DistributedObjectClient> ref = new WeakReference<>(client);
    group.addCallbackOnExitDefaultHandler((state)->{
      DistributedObjectClient ce = ref.get();
      if (ce != null) {
        ce.dump();
        ce.shutdown();
      }
    });
```

ramsai1729:
This change is causing test failures as  `L1ThrowableHandler` keeping the `MessageTransportListener` to be collected (see `ConnectionWatcher`). Maybe using `WeakReference` should work.

myronkscott:
looks like this change needs more thought.  Let's table it for now.  Thanks for catching

myronkscott:
obsolete

PR #1193, opened 2020-10-05
convert NodeID to ServerID

myronkscott:


mathieucarbou (approved):
@tmesic99 : FYI: fix for the passive sync hanging

PR #1195, opened 2020-11-11
PortManager adoption

cljohnso:
This commit stream replaces `PortChooser` with `PortManager` from `terracotta-utilities`.  It's possible that removal of `PortChooser` is too aggressive a move -- this can be reverted and the class marked as deprecated instead if that is what is desired.

After this, a version of `PortChooser` remains in galvan which, at a minimum, needs to be deprecated if not removed entirely.

Sorry for all the whitespace changes -- seems many of the files I touched had trailing whitespace which my editor aggressively removed.  The review is smaller if whitespace changes are ignored.


common/src/test/java/com/tc/net/core/NoReconnectThreadTest.java
```diff
@@ -117,52 +118,64 @@ public void testConnectionEstablisherThreadExit() throws Exception {
...
-      System.err.println(".");
-    }
-
-    // None of the clients should start the ClientConnectionEstablisher Thread for reconnect as the Client
-    // CommsManager is created with reconnect 0. we might need to wait till the created CCE gets quit request.
-    while (getThreadCount(ClientConnectionEstablisher.RECONNECT_THREAD_NAME) > baseAsyncThreads) {
-      ThreadUtil.reallySleep(1000);
-      System.err.println("-");
+    try {
+      listener.start(Collections.emptySet());
```
chrisdennis, line 122: If the 'resource open' is this call then it should be outside the try - no?

common/src/test/java/com/tc/net/core/TCWorkerCommManagerTest.java
```diff
@@ -67,6 +67,7 @@
 import java.util.Properties;
 import java.util.concurrent.Callable;
 import org.junit.Ignore;
+import org.terracotta.utilities.test.net.PortManager;
 
 public class TCWorkerCommManagerTest extends TCTestCase {
```
chrisdennis, line 72: The majority of changes in here appear to be formatting related? Would it be better to stick to the minimum necessary changes here?

chrisdennis (approved):

common/src/test/java/com/tc/net/core/NoReconnectThreadTest.java
cljohnso: You are correct; it seems that the call to `start` is what "opens" the listener -- could be moved out of the try block.

common/src/test/java/com/tc/net/core/TCWorkerCommManagerTest.java
cljohnso: As I noted in the comments for the review, many of the files had trailing whitespace which my editor dutifly removed when the file was saved.  I'm not inclined to put it back.

PR #1197, opened 2020-12-10
Fix retirement of oldest transaction id

myronkscott:
Re-implement oldest transaction id tracking using a sliding window bitset retirement.

### DIFF
--- a/dso-l1/src/main/java/com/tc/object/TransactionSource.java
+++ b/dso-l1/src/main/java/com/tc/object/TransactionSource.java
@@ -19,32 +19,48 @@
 package com.tc.object;
 
 import com.tc.object.tx.TransactionID;
-
-import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.BitSet;
 import java.util.concurrent.atomic.AtomicLong;
 
 public class TransactionSource {
 
-  private static final AtomicLong nextTransactionId = new AtomicLong(1L);
-
-  private static final ConcurrentSkipListSet<Long> pending = new ConcurrentSkipListSet<>();
+  private final AtomicLong current = new AtomicLong();
+  private volatile TransactionID oldestCache = TransactionID.NULL_ID;
+  private long retiredBase = 1L;
+  private int retirePosition = 0;
+  private final long GC_THRESHOLD = 32 * 1024;
+  private BitSet retired = new BitSet();
 
   public TransactionID create() {
-    long txn;
-    while (!pending.add((txn = nextTransactionId.getAndIncrement())));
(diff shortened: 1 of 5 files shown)


dso-l1/src/main/java/com/tc/object/TransactionSource.java
```diff
@@ -19,32 +19,39 @@
...
   }
 
-  public TransactionID oldest() {
-    Long first = pending.first();
-    if (first == null) {
-      return null;
-    } else {
-      return new TransactionID(first);
+  private void gc() {
+    if (retirePosition + retiredBase + (retirePosition>>8) > current.get()) {
```
chrisdennis, line 43: Not sure I follow the logic here. I would have expected the GC logic to be something like:
```java
if (retired.nextClearBit(0) > THRESHOLD) {
 //do gc
}

File: dso-l1/src/main/java/com/tc/object/TransactionSource.java [hunk 2]
```diff
@@ -19,32 +19,39 @@
...
 
   public TransactionID create() {
-    long txn;
-    while (!pending.add((txn = nextTransactionId.getAndIncrement())));
-    return new TransactionID(txn);
+    return new TransactionID(current.incrementAndGet());
+  }
+
+  public synchronized TransactionID oldest() {
+    retirePosition = retired.nextClearBit(retirePosition);
```
chrisdennis, line 37: Wouldn't it be easier to make 'oldest' a volatile and have an unsynced read here (prevent contention on message send), and then update it during the retire call?

chrisdennis (changes requested):

dso-l1/src/main/java/com/tc/object/TransactionSource.java
myronkscott: it's a moving threshold.  The bigger the opportunity to reduce size the greater the chance of doing the work.  Trying to prevent gc while there are a large number of transactions in-flight

dso-l1/src/main/java/com/tc/object/TransactionSource.java
myronkscott: I'm not sure what the read guarantees are for BitSet.  I was erroring on the side of caution.

myronkscott: I see what you are saying now

myronkscott:
PR has been updated.  It is better now.  Retirement is mostly single threaded in the normal case so contention on the synchronization should not be expensive.

chrisdennis (approved):

PR #1209, opened 2021-03-05
Using jcommander for voter script.

mobasherul:


mobasherul:
./start-tc-voter.sh -help
Usage: (-server <hostname:port>,<hostname:port>... -server <hostname:port>,<hostname:port> ... | -override <hostname:port>) [-security-dir <security-dir>]
Options:
    -help         (optional)    Help
    -override     (optional)    Override vote to host:port
    -security-dir (optional)    Security root directory
    -server       (optional)    Servers to connect to

voter/src/main/java/org/terracotta/voter/CustomJCommander.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Override
+    public void usage(StringBuilder out, String indent) {
+      appendOptions(commander, out, indent);
+    }
+
+
+    private void appendOptions(JCommander jCommander, StringBuilder out, String indent) {
+      out.append("Usage: ");
+      String usage = "(-server <hostname:port>,<hostname:port>... -server <hostname:port>,<hostname:port> ... | -override <hostname:port>) [-security-dir <security-dir>]";
```
mathieucarbou, line 66: we should perhaps have the same kind of parameters as other tools. On other tools we use `-connect-to`. Is the `-server` options have the same meaning as `-connect-to` ?

voter/src/main/java/org/terracotta/voter/CustomJCommander.java
```diff
@@ -0,0 +1,110 @@
...
+      out.append(usage.replace(lineSeparator(), lineSeparator() + "    " + indent)).append(lineSeparator());
+      // Align the descriptions at the "longestName" column
+      List<ParameterDescription> sorted = Lists.newArrayList();
+
+      int colSize = "-security-dir".length(); // quick hack to get the largest possible option
```
mathieucarbou, line 71: should be `-security-root-directory` like other tools

voter/src/main/java/org/terracotta/voter/CustomJCommander.java
mathieucarbou, line 66: I would change `-override` to something more meaningful. What it overrides exactly ?

voter/src/main/java/org/terracotta/voter/options/OptionsParsingImpl.java
```diff
@@ -0,0 +1,62 @@
...
+  @Parameter(names = {"-help", "-h"}, description = "Help", help = true)
+  private boolean help;
+
+  @Parameter(names = {"-override", "-o"}, description = "Override vote to host:port")
+  private String overrideHostPort;
+
+  @Parameter(names = {"-server", "-s"}, description = "Comma separated host:port to connect to (one per stripe)", splitter = NoCommaSplitter.class)
+  private List<String> serversHostPort;
+
+  @Parameter(names = {"-security-dir", "-srd"}, description = "Security root directory")
```
mathieucarbou, line 39: same remarks here as I did before for:
- `-override` => `-override-<something>`
- `-server` => `-connect-to`
- `-security-dir` => `-security-root-directory`

voter/src/main/java/org/terracotta/voter/CustomJCommander.java
mobasherul: I kept it as security-dir as it is used in server startup script

voter/src/main/java/org/terracotta/voter/CustomJCommander.java
mobasherul: It gives a vote to the host:port passed

mobasherul: I will change it to connect-to

voter/src/main/java/org/terracotta/voter/CustomJCommander.java
mathieucarbou: yes but it can be hidden and we could just show this new option only ion the help, like we did for the other tools.
Goal is to aligne all the tools and helps, and still keep backward compatibility.
I would also update the startup scripts because we have a control over them.

voter/src/main/java/org/terracotta/voter/CustomJCommander.java
mathieucarbou: `-override-vote` then ;-)

mathieucarbou: or if it gives a vote... `-vote-for` instead of `-override`

myronkscott, line 66: -security-dir does not belong here.  This is OSS

voter/src/main/java/org/terracotta/voter/CustomJCommander.java
myronkscott, line 71: let's try and avoid leaking EE features in an OSS project.

voter/src/main/java/org/terracotta/voter/options/Options.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  public void setServerHostPort(List<String> serversHostPort) {
+    this.serversHostPort = serversHostPort;
+  }
+
+  public void setSecurityDir(String securityDir) {
+    this.securityDir = securityDir;
+  }
+
```
myronkscott, line 45: EE leak

voter/src/main/java/org/terracotta/voter/options/Options.java
```diff
@@ -0,0 +1,61 @@
...
+    return overrideHostPort;
+  }
+
+  public List<String> getServersHostPort() {
+    return serversHostPort;
+  }
+
+  public String getSecurityDir() {
+    return securityDir;
+  }
```
myronkscott, line 60: EE leak

myronkscott (changes requested):
Mainly concerned about leaking EE features here in an OSS project.  I believe there are examples elsewhere that make this pluggable so options can live next to implementation

mathieucarbou (approved):

mathieucarbou (approved):

mathieucarbou (approved):

mathieucarbou:
@myronkscott : Can you please have another look. Thanks!

myronkscott (approved):

PR #1222, opened 2021-05-25
update kit so it can be used more efficiently downstream

myronkscott:
@akomakom the init script is not longer relevant.  Should we remove it?  Actually, I think we should just remove the arguments in terracotta.properties


akomakom:
> @akomakom the init script is not longer relevant. Should we remove it? Actually, I think we should just remove the arguments in terracotta.properties

@myronkscott are you referring to https://github.com/Terracotta-OSS/terracotta-core/blob/master/terracotta-kit/src/assemble/init/sysv/terracotta.properties#L38 ?

If the init script isn't used, then both files should be removed.  What changed?  I don't have context.

chrisdennis (approved):
I think this makes sense. I assume you tested the integration of this with the downstream tc--ee repo?

myronkscott:
tested in EE.  Merging

PR #1223, opened 2021-06-14
fix group manager handshake serialization

myronkscott:


tc-server/src/main/java/com/tc/net/groups/TCGroupManagerImpl.java
```diff
@@ -1076,37 +1078,27 @@ protected void switchToState(HandshakeState state) {
...
-        waitForStateTransitionToComplete();
-        stateTransitionInProgress = true;
+        this.current = next;
       }
+      
       if (isDebugLogging()) {
         debugInfo("[TCGroupHandshakeStateMachine]: Entering state: " + state + ", for channel: " + channel);
       }
+      previous.waitForCompletion();
+      next.start();
```
chrisdennis, line 1099: Could we embed the `state.enter()` call inside the monitor and make it:

```
previous.waitForComplete()
next.enter()
```

Where `next.enter()` would "lock", call and then "unlock"?

chrisdennis (approved):
Approved... I think this works. One comment you might or might not want to act upon.

tc-server/src/main/java/com/tc/net/groups/TCGroupManagerImpl.java
myronkscott: good suggestion.

PR #1230, opened 2021-06-25
Gather and display information after initialization

mathieucarbou:
Related to PR https://github.com/Terracotta-OSS/terracotta-platform/pull/1019

And discussion: https://github.com/Terracotta-OSS/terracotta-platform/pull/1019#issuecomment-868115092

### DIFF
--- a/tc-server/src/main/java/com/tc/server/Bootstrap.java
+++ b/tc-server/src/main/java/com/tc/server/Bootstrap.java
@@ -102,7 +102,7 @@ public class Bootstrap implements BootstrapService {
     String longProductString = info.toLongString();
     CONSOLE.info(longProductString);
 
-    CONSOLE.info("Extensions:");
+    CONSOLE.info("Installed extensions:");
     for (String ext : info.getExtensions()) {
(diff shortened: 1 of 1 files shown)


tc-server/src/main/java/com/tc/server/Bootstrap.java
```diff
@@ -103,7 +102,7 @@ private static void writeVersion(ProductInfo info) {
     String longProductString = info.toLongString();
     CONSOLE.info(longProductString);
 
-    CONSOLE.info("Extensions:");
+    CONSOLE.info("Installed extensions:");
```
mathieucarbou, line 105: I am also proposing this change because the logs are displaying **ALL** installed extensions through meta-inf/services mechanism. This is not reflecting which extension is actually used at runtime. I.e. some extensions could be disabled through configuration or diagnostic mode.

myronkscott:
I think I have a better way to deal with this issue. Please make sure that help goes to stdout or stderr and the issue should be fixed I. The next release.

PR #1231, opened 2021-06-28
miscellaneous fixes for test execution

myronkscott:


common/src/main/java/com/tc/lang/TCThreadGroup.java
```diff
@@ -105,6 +106,24 @@ public boolean retire(long timeout, Consumer<InterruptedException> interruptHand
     return complete;
   }
 
+  private boolean lookForThreadExit(Thread t, Consumer<InterruptedException> interruptHandler) {
+    if (ignorePoolThreads && (t.getName().startsWith("pool-") || t.getName().startsWith("ForkJoinPool.commonPool-worker-"))) {
```
chrisdennis, line 110: `t instanceof ForkJoinWorkerThread` would at least make this slightly less ugly.

common/src/main/java/com/tc/util/version/VersionCompatibility.java
```diff
@@ -47,6 +47,20 @@ private static boolean isCompatible(Version v1, Version v2) {
     return ((v1.major() == v2.major()) && (v1.minor() == v2.minor()));
   }
 
+  public static boolean isNewer(Version v1, Version v2) {
+    if (v1 == null || v2 == null) { throw new NullPointerException(); }
+    if (v1.major() > v2.major()) {
```
chrisdennis, line 52: `return v1.compareTo(v2) > 0`

chrisdennis (changes requested):

common/src/main/java/com/tc/lang/TCThreadGroup.java
myronkscott: that would mean any fork join worker thread.  not just the common pool.

common/src/main/java/com/tc/util/version/VersionCompatibility.java
myronkscott: that goes very deep including patch and build number.  I don't think we want to go that deep with this.

common/src/main/java/com/tc/lang/TCThreadGroup.java
chrisdennis: Sorry...:
```
(t instanceof ForkJoinWorkerThread) && ((ForkJoinWorkerThread) t).getPool() == ForkJoinPool.commonPool()
```

chrisdennis:
@myronkscott I approved... let me know if you want me to merge.

PR #1234, opened 2021-07-21
Add client revision to handshake info and TCServerInfo mbean.

tmesic99:
Connected client versioning information, (e.g. version and revision), needs to be retrievable by the Diagnostic Service.  The current changes include (1) ability for connecting clients to provide their build revision string as part of the connection handshake and (2) the client version details exposed via the existing DSO MBean.,

### DIFF
--- a/tc-server/src/main/java/com/tc/stats/api/DSOMBean.java
+++ b/tc-server/src/main/java/com/tc/stats/api/DSOMBean.java
@@ -19,7 +19,9 @@
 package com.tc.stats.api;
 
 import com.tc.management.TerracottaMBean;
+import com.tc.stats.Client;
 
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
@@ -37,6 +39,8 @@ public interface DSOMBean extends TerracottaMBean {
   static final String CLIENT_DETACHED = "dso.client.detached";
 
   ObjectName[] getClients();
+
+  List<Client> getConnectedClients();
 
   Map<ObjectName, Exception> setAttribute(Set<ObjectName> onSet, String attrName, Object attrValue);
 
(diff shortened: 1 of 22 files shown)


common/src/main/java/com/tc/util/ProductInfo.java
```diff
@@ -240,10 +244,6 @@ public String copyright() {
     return buildInfo.getCopyright();
   }
 
-  public String buildRevision() {
-    return buildInfo.getRevision();
-  }
-
```
tmesic99, line 246: didn't really mean to relocate this existing method in the file.

myronkscott:
You are changing the handshake message which will impact client <-> compatibility between versions.  You will need to come up with a scheme where an old client version can talk to a new server.

myronkscott:
just taking a step back, having the build revision not unique to the stated version would be strange.  What is the use-case for having the build info published but the client?

tc-server/src/main/java/com/tc/stats/api/DSOMBean.java
```diff
@@ -38,6 +39,8 @@
 
   ObjectName[] getClients();
 
+  String getClientsInformation() throws IOException;
```
myronkscott, line 42: Instead of adding this method which seems to be single use, it might be better to add the version info to the Client MBean and iterate with the results from getClients()

tmesic99:
> 
> 
> You are changing the handshake message which will impact client <-> compatibility between versions. You will need to come up with a scheme where an old client version can talk to a new server.

These code changes support any older 10.7.x client connecting to a server running with these new changes.  For example, clients from our 10.7.0 general release plus clients running the 10.7 fixes 1-3 can all connect to this new server without the revision being supplied in the handshake.  
Note that 10.3 and 10.5 clients can not connected to 10.7.  Not because of this new revision attribute, but for a different reason:

Exception: org.terracotta.exception.ConnectionClosedException: 
Entity: com.terracottatech.tools.client.TopologyEntity:topology-entity 
Connection closed before sending message
Exception: java.lang.NullPointerException.

Chris/James confirmed that 10.5 clients do not work with 10.7 servers.

tc-server/src/main/java/com/tc/stats/api/DSOMBean.java
tmesic99: Added getVersion/Revision to the Client MBean, deriving values from minfo.  
Added getConnectedClients() to DSO MBean to access all clients' details.
Moved getClientsInformation to TCServerInfo.

myronkscott (approved):

PR #1237, opened 2021-08-09
Reconnect change

myronkscott:
During reconnect the low-level tc-connection from the client is inserted into a pre-prepared network stack.  If the connection fails to complete handshake or the client abandons the connection attempt, the server tears down the network stack and the client will never connect.  This change will allow a client to restart the process during the reconnect window and attempt to make and bootstrap a successful connection.

### DIFF
--- a/tc-server/src/test/java/com/tc/net/protocol/transport/ServerStackProviderTest.java
+++ b/tc-server/src/test/java/com/tc/net/protocol/transport/ServerStackProviderTest.java
@@ -113,10 +113,8 @@ public class ServerStackProviderTest extends TCTestCase {
     }
 
     // Now make the attach fail due to IllegalReconnectException, we should force a reconnection rejected exception now
-//    when(harness.attachNewConnection(any(TCConnection.class))).thenThrow(new IllegalReconnectException());
     try {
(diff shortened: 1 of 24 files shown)


tc-server/src/test/java/com/tc/net/protocol/transport/ServerStackProviderTest.java
```diff
@@ -113,10 +113,8 @@ public void testAttachAlreadyConnectedTransport() throws Exception {
     }
 
     // Now make the attach fail due to IllegalReconnectException, we should force a reconnection rejected exception now
-//    when(harness.attachNewConnection(any(TCConnection.class))).thenThrow(new IllegalReconnectException());
     try {
       serverStackProvider.attachNewConnection(nextConnectionID, connection);
```
cljohnso, line 117: Does this part of the test have any continued relevance?

cljohnso (approved):

tc-server/src/test/java/com/tc/net/protocol/transport/ServerStackProviderTest.java
myronkscott: not really

PR #1243, opened 2021-09-22
fix canDelete logic when creating entities

tmesic99:
For this change, the logic for determining if an entity can be deleted is based on whether or not the entity is a PermanentEntity.  The previous approach determined an entity's deletion status based on whether or not the entity creation requestor was a client.  That approach did not work for dynamic scaling whereby entities can be created by non-clients (i.e. the server itself).

### DIFF
--- a/tc-server/src/main/java/com/tc/objectserver/entity/ServerEntityFactory.java
+++ b/tc-server/src/main/java/com/tc/objectserver/entity/ServerEntityFactory.java
@@ -52,12 +52,14 @@ import org.terracotta.entity.PassiveServerEntity;
 import org.terracotta.entity.EntityServerService;
 
 import java.util.List;
+import java.util.function.Predicate;
 
 /**
  * @author twu
  */
 public class ServerEntityFactory {
   private final ServiceLocator locator;
+  private List<VoltronEntityMessage> entityMessages;
   
   public ServerEntityFactory(ServiceLocator loader) {
     this.locator = loader;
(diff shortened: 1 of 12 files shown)


tc-server/src/main/java/com/tc/objectserver/entity/ServerEntityFactory.java
```diff
@@ -92,7 +93,15 @@ public ServerEntityFactory(ServiceLocator loader) {
...
-  }  
+  }
+
+  public boolean isPermanentEntity(EntityID entity) {
+    Predicate<EntityID> sameName = e -> e.getEntityName().equals(entity.getEntityName());
+    Predicate<EntityID> sameClass = e -> e.getClassName().equals(entity.getClassName());
+    return getAnnotatedEntities().stream()
+        .map(p -> p.getEntityDescriptor().getEntityID())
+        .anyMatch(sameName.and(sameClass));
+  }
```
myronkscott, line 105: should getAnnotatedEntities() be built once and held in a map/list since it is now being queried on every entity create?

tc-server/src/main/java/com/tc/objectserver/persistence/EntityPersistor.java
```diff
@@ -161,7 +161,6 @@ public synchronized void entityCreateFailed(EntityID eid, ClientID clientID, lon
   public synchronized void entityCreated(ClientID clientID, long transactionID, long oldestTransactionOnClient, EntityID id, long version, long consumerID, boolean canDelete, byte[] configuration) {
     LOGGER.debug("entityCreated " + clientID + " " + transactionID + " " + id + " " + version);
     Assert.assertTrue(canDelete);
-    Assert.assertFalse(clientID.isNull());
```
myronkscott, line 164: this should be reverted

myronkscott: See ProcessTransactionHandler.java:923.  That, I think, needs to check isPermanentEntity again

myronkscott (changes requested):

tc-server/src/main/java/com/tc/objectserver/handler/ProcessTransactionHandler.java
```diff
@@ -388,7 +388,8 @@ private void addMessage(ClientID sourceNodeID, EntityDescriptor descriptor, Serv
       capture.setTransactionOrderPersistenceFuture(transactionOrderPersistenceFuture);
       try {
         EntityID entityID = descriptor.getEntityID();
-        ManagedEntity temp = entityManager.createEntity(entityID, descriptor.getClientSideVersion(), consumerID, !capture.getNodeID().isNull());
+        boolean canDelete = !entityManager.getEntityLoader().isPermanentEntity(entityID);
+        ManagedEntity temp = entityManager.createEntity(entityID, descriptor.getClientSideVersion(), consumerID, canDelete);
```
chrisdennis, line 392: Given that the `EntityManager` is going to call the `EntityServerService` to create the entity, wouldn't it be easier to just remove the boolean parameter entirely and let the EntityManager implementation ask the server service either via a new method, or by checking for the annotations whether a given construction should be permanent or not?

tc-server/src/main/java/com/tc/objectserver/entity/ServerEntityFactory.java
tmesic99: Is the complete set of AnnotatedEntities populated only once (e.g. at server startup)?  Or do additional ones get added during a server's lifetime?

tc-server/src/main/java/com/tc/objectserver/handler/ProcessTransactionHandler.java
tmesic99: Sounds good.  What is our policy about changing interface method signatures (to remove the boolean)?

tc-server/src/main/java/com/tc/objectserver/entity/ServerEntityFactory.java
myronkscott: It's a classpath scan so it should not change in the life of the server.

tc-server/src/main/java/com/tc/objectserver/handler/ProcessTransactionHandler.java
myronkscott: interface face changes are fine as long as they are contained in tc-core.

myronkscott: one thing to consider that while permanent and temporary entities are an annotation driven concept, they didn't start that way.  Right now, temporary and permanent entity concepts are not consistent with each other.  Permanent entities have a runtime aspect to them while temporary entities are annotation only.  We probably should decide which way to go and make them consistent with each other.

tc-server/src/main/java/com/tc/objectserver/entity/ServerEntityFactory.java
tmesic99: done

tc-server/src/main/java/com/tc/objectserver/persistence/EntityPersistor.java
tmesic99: done

tc-server/src/main/java/com/tc/objectserver/handler/ProcessTransactionHandler.java
tmesic99: > 
> 
> Given that the `EntityManager` is going to call the `EntityServerService` to create the entity, wouldn't it be easier to just remove the boolean parameter entirely and let the `EntityManager` implementation ask the server service either via a new method, or by checking for the annotations whether a given construction should be permanent or not?

done

myronkscott (approved):

PR #1248, opened 2021-11-09
retry on DNS problems

myronkscott:


chrisdennis (changes requested):
This makes sense. Any chance we could get some tests for the half duplex detection logic?

myronkscott:
Obsolete.  Will create a new pull request for unmerged items.

PR #1250, opened 2021-12-16
Remove recycle in comms messaging

myronkscott:


myronkscott:
also narrowing greatly the scope of buffer usage eliminating direct buffers for the time being.

PR #1251, opened 2022-01-27
TDB-3830 better feedback for voter tool

tmesic99:
Implements better voter tool feedback as described in TDB-3830

### DIFF
--- a/voter/src/main/java/org/terracotta/voter/ActiveVoter.java
+++ b/voter/src/main/java/org/terracotta/voter/ActiveVoter.java
@@ -170,7 +170,15 @@ public class ActiveVoter implements AutoCloseable {
             if (response >= 0) {
               registrationLatch.complete(voterManager);
             } else {
-              LOGGER.warn("Registration with {} in state {} failed. Retrying...", voterManager.getTargetHostPort(), voterManager.getServerState());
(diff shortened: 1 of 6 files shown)


voter/src/main/java/org/terracotta/voter/ActiveVoter.java
```diff
@@ -170,7 +170,15 @@ ClientVoterManager registerWithActive(String id,
             if (response >= 0) {
               registrationLatch.complete(voterManager);
             } else {
-              LOGGER.warn("Registration with {} in state {} failed. Retrying...", voterManager.getTargetHostPort(), voterManager.getServerState());
+              StringBuilder message = new StringBuilder();
+              message.append(String.format("Registration with %s in state %s failed. ",
+                  voterManager.getTargetHostPort(), voterManager.getServerState()));
+              long voterLimit = Math.max(0, voterManager.getRegisteredVoterLimit());
```
tmesic99, line 176: When failover priority is set to 'availability', getRegisteredVoterLimit() returns -1.  So I just use the max of '0' to generate the message that "Configured voter limit (0) has been reached"..  Alternatively, I could have prepared a dedicated message just for the -1 case... something like, "voters are not supported when failover priority is set to availability".

PR #1261, opened 2022-04-07
Move voter tool cli from core into platform

tmesic99:
Remove the voter tool cli code out of core and into platform (along side other client cli tools).
This work is captured in WF-15979.


myronkscott (approved):

myronkscott:
please rebase on top of main

PR #1273, opened 2022-06-27
Remove SetOnceFlag and SetOnceRef and replace with appropriate AtomicBoolean and AtomicReference usage

chrisdennis:


myronkscott:
Why do you want to get rid of these?  I've used them in the past for things for flags the go positive and never flip back.

chrisdennis:
I guess my first thought was not to remove them, but to reimplement them without synchronized blocks (particularly the one on read for `SetOnceRef`). Then I figured the actual used code was pretty minimal so I figured it might be simpler to just remove them. On review some of the usages actually just dropped away to volatile boolean fields.

PR #1274, opened 2022-07-08
Use version range for logging components & terracotta-utilities

cljohnso:
This commit introduces version ranges for Logback and SLF4J.

This commit also introduces a version range for terracotta-utilities
and bumps the base version.  The upper bound is left open.

To prevent depdendency convergence conflicts, several exclusions
(mostly for SLF4J) are also included.

### DIFF
--- a/galvan-support/pom.xml
+++ b/galvan-support/pom.xml
@@ -57,6 +57,12 @@ limitations under the License.
       <groupId>org.terracotta</groupId>
       <artifactId>galvan</artifactId>
       <version>${galvan.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.terracotta</groupId>
+          <artifactId>terracotta-utilities-port-chooser</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.terracotta</groupId>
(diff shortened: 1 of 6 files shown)


galvan-support/pom.xml
```diff
@@ -57,6 +57,12 @@ limitations under the License.
       <groupId>org.terracotta</groupId>
       <artifactId>galvan</artifactId>
       <version>${galvan.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.terracotta</groupId>
+          <artifactId>terracotta-utilities-port-chooser</artifactId>
```
chrisdennis, line 63: Is this temporary until Galvan gets a new release?

tc-client/pom.xml
```diff
@@ -43,6 +43,10 @@
   </properties>
 
   <dependencies>
+    <dependency>
```
chrisdennis, line 46: This seems odd? Is it missing due to transitive exclusion now?

cljohnso, line 46: I'm not quite sure from where this module was getting its slf4j-api library but, with the other changes, this module fails to compile without this dependency declaration.  (Code in this module has direct use of SLF4J so a direct declaration is not out of place.)

cljohnso:
I still need to verify that the "kit" content is OK ...

galvan-support/pom.xml
cljohnso: No, it is not.  This module has a direct dependency on port-chooser and Maven's busted dependency convergence enforcement pulls in the base version declared by galvan instead of the version actually used or resolved in this module.  Icky but necessary.

tc-client/pom.xml
cljohnso: You were quick ... I addressed this below.

cljohnso:
> I still need to verify that the "kit" content is OK ...

The content of `target/terracotta-5.9-SNAPSHOT/terracotta-5.9-SNAPSHOT` appears to be consistent with the version built from master:HEAD along with the expected version differences.

mathieucarbou (approved):

myronkscott (approved):

PR #1276, opened 2022-07-11
[daggy] Reduce log level and fixing message

mathieucarbou:


tc-client/src/main/java/com/tc/object/DistributedObjectClient.java
```diff
@@ -386,13 +386,13 @@ private void openChannel(ClientMessageChannel channel) throws InterruptedExcepti
         DSO_LOGGER.debug("Channel open");
         break;
       } catch (final TCTimeoutException tcte) {
-        DSO_LOGGER.info("Unable to connect to server/s {} ...sleeping for 5 sec.", serverAddresses);
+        DSO_LOGGER.debug("Unable to connect to server/s {} ...sleeping for 1 sec.", serverAddresses);
```
myronkscott, line 389: I would just remove if you are changing log level.  The message under debug suffices.  I think the user will need to know however that the client is trying to connect and is failing.

PR #1279, opened 2022-07-25
Kit cleanup

myronkscott:
Cleanup unused cruft in the kit.  provide a mechanism to start an extended server.


mathieucarbou:
> the way to start an extended server is to create a new server jar that contains the extensions (logging config, extra log classpath, etc.) and a classpath reference to tc-server.jar. then point the bootstrapping process to the new server using the system property tc.server-jar (along with tc.install-root which is already provided in the script).

question about this new jar to create: how to define its classpath ?

- only the added logback libs + tc-server.jar is enough ? (because the jvm will be able to see when loading the tc-server.jar that it needs to inclue its own dependencies)
- or we need to create a jar with a dependency set which is the union of the tc-server.jar dependencies + our own jar dependencies ?

In case this is the second point (and the "entrypoint" jar needs to know about the whole classpath), how do we depend on tc-server.jar plus all its dependencies ? There is a pom coordinate for it ?

mathieucarbou (approved):
thanks! I just have some questions regarding how to make that work ;-)

myronkscott:
you should be able to just include tc-server.jar + additions.  class loader will pick up additional classpath items in the manifest of tc-server.jar

PR #1281, opened 2022-08-12
Async client implementation

chrisdennis:


galvan-support/src/test/java/org/terracotta/testing/rules/ClientLeakIT.java
```diff
@@ -41,6 +42,7 @@ public class ClientLeakIT {
   /**
    * This will ensure that a fail-over correctly happens.
    */
+  @Ignore
```
myronkscott, line 45: What's happening with this test?

tc-client/src/main/java/com/tc/client/ClientFactory.java
```diff
@@ -30,7 +30,7 @@ public class ClientFactory {
   //  we can verify that it won't be used here.
   public static DistributedObjectClient createClient(Iterable<InetSocketAddress> serverAddresses, ClientBuilder builder,
                                                      TCThreadGroup threadGroup,
-                                                     String uuid, String name, boolean async) {
-    return new DistributedObjectClient(serverAddresses, builder, threadGroup, uuid, name, async);
+                                                     String uuid, String name) {
+    return new DistributedObjectClient(serverAddresses, builder, threadGroup, uuid, name);
```
myronkscott, line 34: I'd leave this with the BasicConnectionManager for diagnostic connections.  Basic has been stable for a while and I think synchronous is way better for those cases.

tc-client/src/main/java/com/tc/object/ClientEntityManagerImpl.java
```diff
@@ -93,106 +101,39 @@ public class ClientEntityManagerImpl implements ClientEntityManager {
...
-    return interrupted;
+  @Override
+  public synchronized boolean isValid() {
+    return !stateManager.isShutdown() && channel().map(ClientMessageChannel::isOpen).orElse(false);
   }
 
   private boolean enqueueMessage(InFlightMessage msg) throws RejectedExecutionException {
-    if (!this.stateManager.isRunning()) {
+    if (this.stateManager.isShutdown()) {
       return false;
```
myronkscott, line 130: so if the connection is paused now the message is still enqueued?  Seems to make sense but want to confirm this situation is covered under async api

chrisdennis: Yep. Since we can't block on enqueue without breaking the async-ness we have to take all comers, they get queued, and then will get sent with the handshake message when we unpause.

myronkscott (approved):
LGTM.  I'd like to keep synchronous connections for diagnostic but if it makes things too complicated we are only talking about one extra thread.

chrisdennis:
> I'd like to keep synchronous connections for diagnostic

That makes sense. Let me have a crack at that today/tomorrow and see if it can be done.

common/src/main/java/com/tc/net/protocol/transport/ClientMessageTransport.java
```diff
@@ -305,9 +305,7 @@ private void handleSynAck(WireProtocolMessage message) {
           handleHandshakeError(new TransportHandshakeErrorContext(synAck.getErrorContext() + " " + message,
                                                                   synAck.getErrorType()));
         }
-      }
-
-      if (!getConnectionID().isNewConnection() && getConnectionID().isValid()) {
+      } else if (!getConnectionID().isNewConnection() && getConnectionID().isValid()) {
```
myronkscott, line 308: This worries me a little.  Even if there is a handshake error on reconnect, shouldn't the server not be treating this as a new connection?

myronkscott: or is the connection id not properly populated on error

tc-client/src/main/java/com/tc/object/DistributedObjectClient.java
```diff
@@ -264,7 +256,9 @@ private synchronized ClientMessageChannel internalStart(int socketTimeout) {
                                                                  socketTimeout);
 
 
-    final ClientIDLoggerProvider cidLoggerProvider = new ClientIDLoggerProvider(clientChannel::getClientID);
+    WeakReference<ClientMessageChannel> channelRef = new WeakReference<>(clientChannel);
+
```
myronkscott, line 260: Yuck...should we just get rid of the special logger?  Or maybe use logback or slf4j conventions to achieve something similar?

myronkscott (approved):

common/src/main/java/com/tc/net/protocol/transport/ClientMessageTransport.java
chrisdennis: Yes probably. My experimentation say this is a pre-existing problem. It's masked in the existing BasicConnection implementation because the thrown AssertionError is in a non TCThreadGroup thread, so it doesn't kill the client. I think the BasicConnection just fires another attempt at the readExecutor and continues on.

PR #1283, opened 2022-09-07
Provide protection from untrusted callbacks when calling user code

chrisdennis:
This needs test coverage... but I think this kind of approach makes sense. If you're happy enough with the principle here, then I'll add some test coverage for the guard class and we can merge.

### DIFF
--- a/tc-client/src/main/java/com/tc/object/BinaryInvocationCallback.java
+++ b/tc-client/src/main/java/com/tc/object/BinaryInvocationCallback.java
@@ -19,15 +19,14 @@
 package com.tc.object;
 
 import org.terracotta.entity.EntityResponse;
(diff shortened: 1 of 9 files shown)


myronkscott:
Makes sense.  There is a danger that since the client (user code) is responsible for setting up logging that these can get swallowed silently.  I don't think they would happen normally but I don't have a full conception of who will be crafting these callbacks.  Maybe this is good for now and we see what happens?

chrisdennis:
These can eventually end up executing user code, and that's where I expect the majority of these exceptions to come from. To that end the most important thing is to protect the client by catching the exceptions. The logging is just us being polite about it.

myronkscott (approved):

PR #1284, opened 2022-09-08
Reference tracking reduction

myronkscott:


common/src/main/java/com/tc/async/api/SEDA.java
```diff
@@ -48,25 +48,4 @@ public StageManager getStageManager() {
   protected TCThreadGroup getThreadGroup() {
     return this.threadGroup;
   }
-  
-  public void stageWarning(Object description) {
-    
-  }
-  
-  private static class StallWarning {
```
chrisdennis, line 56: Is this change unrelated?

common/src/main/java/com/tc/net/protocol/tcm/CommunicationsManagerImpl.java
```diff
@@ -175,8 +177,11 @@ public CommunicationsManagerImpl(MessageMonitor monitor, TCMessageRouter message
     this.connectionManager = connMgr;
 
     Assert.eval(healthCheckerConfig != null);
+    ReferenceQueue<Object> gc = new ReferenceQueue<>();
+    PhantomReference<Object> ref = new PhantomReference<>(this, gc);
```
chrisdennis, line 181: From memory phantom references are more expensive during GC than weak ones. Not that there'll be many of them... but does this need to be phantom?

common/src/main/java/com/tc/net/protocol/tcm/CommunicationsManagerImpl.java
```diff
@@ -175,8 +177,11 @@ public CommunicationsManagerImpl(MessageMonitor monitor, TCMessageRouter message
     this.connectionManager = connMgr;
 
     Assert.eval(healthCheckerConfig != null);
+    ReferenceQueue<Object> gc = new ReferenceQueue<>();
```
chrisdennis, line 180: This logic could go inside the if condition, right?

tc-client/src/main/java/com/tc/object/DistributedObjectClient.java
```diff
@@ -473,13 +428,13 @@ public ClientEntityManager getEntityManager() {
...
   public String getClientState() {
     PrettyPrinter printer = new MapListPrettyPrint();
-    this.communicationStageManager.prettyPrint(printer);
     this.clientEntityManager.prettyPrint(printer);
     return printer.toString();
   }
 
   public void dump() {
     DSO_LOGGER.info(getClientState());
+    DSO_LOGGER.info(ThreadDumpUtil.getThreadDump());
```
chrisdennis, line 437: Was this just for debugging... or is it compensating for losing the `communicationsStageManager`?

chrisdennis (changes requested):

common/src/main/java/com/tc/async/api/SEDA.java
myronkscott: Cleanup of dead code

myronkscott: was holding references not needed

common/src/main/java/com/tc/net/protocol/tcm/CommunicationsManagerImpl.java
myronkscott: I was under the impression that phantom references where a replacement for finalization which this is basically an implementation of.

tc-client/src/main/java/com/tc/object/DistributedObjectClient.java
myronkscott: It was doing a Thread.dumpStack previously which my IDE doesn't like.

chrisdennis (approved):

PR #1286, opened 2022-10-04
Client health check changes

myronkscott:
Add health-check to diagnostic connections
Move to a single health-check executor per classloader

### DIFF
--- a/common/src/main/java/com/tc/net/protocol/transport/ConnectionHealthCheckerImpl.java
+++ b/common/src/main/java/com/tc/net/protocol/transport/ConnectionHealthCheckerImpl.java
@@ -18,20 +18,22 @@
  */
 package com.tc.net.protocol.transport;
 
+import com.tc.lang.TCThreadGroup;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.tc.net.core.TCConnection;
 import com.tc.net.core.TCConnectionManager;
 import com.tc.util.Assert;
-import com.tc.util.concurrent.SetOnceFlag;
 import java.net.InetSocketAddress;
 
 import java.util.Iterator;
-import java.util.Timer;
-import java.util.TimerTask;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/tc/net/protocol/transport/ConnectionHealthCheckerImpl.java
```diff
@@ -44,18 +46,24 @@
 public class ConnectionHealthCheckerImpl implements ConnectionHealthChecker {
 
   private final Logger logger;
-  private final Timer                           monitorThread;
+  private static final ScheduledExecutorService monitorThread = Executors.newSingleThreadScheduledExecutor((r) -> {
+    ThreadGroup grp = Thread.currentThread().getThreadGroup();
+    while (grp instanceof TCThreadGroup) {
+      grp = grp.getParent();
```
mathieucarbou, line 52: why it cannot be in a TcThreadGroup ?

common/src/main/java/com/tc/net/protocol/transport/ConnectionHealthCheckerImpl.java
```diff
@@ -91,10 +99,9 @@ public void start() {
...
 
   @Override
-  public void stop() {
-    if (shutdown.attemptSet()) {
-      monitorThreadEngine.stop();
-      monitorThread.cancel();
+  public synchronized void stop() {
+    if (task != null && !task.isCancelled()) {
+      task.cancel(true);
 // don't bother to join the monitorThread here.  shutdown should take care of all the
```
mathieucarbou, line 105: is this comment relevant ? the scheduled thread is not in the TCThreadGroup.

common/src/main/java/com/tc/net/protocol/transport/ConnectionHealthCheckerImpl.java
myronkscott: because the client shutdown code will hang on threads that it believes belong to it.  This thread no longer belongs to a specific connection but all the connections

common/src/main/java/com/tc/net/protocol/transport/ConnectionHealthCheckerImpl.java
myronkscott: yeah.  probably not relevant any longer.

common/src/main/java/com/tc/net/protocol/transport/ConnectionHealthCheckerImpl.java
```diff
@@ -44,42 +46,59 @@
...
   private HealthCheckerMonitorThreadEngine getHealthMonitorThreadEngine(HealthCheckerConfig config,
                                                                           TCConnectionManager connectionManager,
                                                                           Supplier<Boolean> reachable,
                                                                           Logger logger) {
-    return new HealthCheckerMonitorThreadEngine(config, connectionManager, ()-> {
-      if (reachable.get()) {
-        return true;
-      } else {
+    Supplier<Boolean> local = () -> {
+      if (!reachable.get()) {
```
chrisdennis, line 84: If you invert this if condition and then inline the Runnable again the changeset is much smaller.

common/src/main/java/com/tc/net/protocol/transport/ConnectionHealthCheckerImpl.java
```diff
@@ -91,22 +110,15 @@ public void start() {
...
 
   @Override
-  public void stop() {
-    if (shutdown.attemptSet()) {
-      monitorThreadEngine.stop();
-      monitorThread.cancel();
-// don't bother to join the monitorThread here.  shutdown should take care of all the
-// threads in the thread group
+  public synchronized void stop() {
+    if (task != null && !task.isCancelled()) {
```
chrisdennis, line 114: This might be better as `!task.isDone()`.

chrisdennis (approved):
Couple of comments you could act on (or not) at your discretion.

common/src/main/java/com/tc/net/protocol/transport/ConnectionHealthCheckerImpl.java
myronkscott: I don't want to re-issue the cancel if it has already been initiated.

PR #1297, opened 2022-10-31
Fix the root level of logging

myronkscott:


myronkscott:
added a small commit for a debugging tool

client-runtime/src/main/java/com/tc/cmdline/Probe.java
```diff
@@ -0,0 +1,45 @@
+/*
```
chrisdennis, line 1: Was this an accidental pickup?

myronkscott:
Separate commit for a debugging tool that I want to get in.
________________________________

myronkscott:
obsolete

PR #1298, opened 2022-11-07
IPv6 fix

mathieucarbou:
Implementors of `NetworkTranslator` can receive a badly formed host:port in case host in an IPv6 address.

See the ugly hack we had to do in DC a the moment to work around that: 

https://github.com/Terracotta-OSS/terracotta-platform/blob/master/dynamic-config/server/services/src/main/java/org/terracotta/dynamic_config/server/service/DynamicConfigNetworkTranslator.java#L70-L79

### DIFF
--- a/tc-server/src/main/java/com/tc/net/groups/Node.java
+++ b/tc-server/src/main/java/com/tc/net/groups/Node.java
@@ -86,7 +86,9 @@ public class Node {
   }
 
   private String getName() {
-    return host + DELIMITER + port;
+    // The node name is passed when constructing a new ServerID()
+    // Other callers of new ServerID() are using getStringForm#getStringForm()
+    return (host.contains(":") ? ("[" + host + "]") : host) + DELIMITER + port;
   }
 
   private static String getHost(String hostPort) {


tc-server/src/main/java/com/tc/net/groups/Node.java
```diff
@@ -86,7 +86,9 @@ public String getServerNodeName() {
   }
 
   private String getName() {
-    return host + DELIMITER + port;
+    // The node name is passed when constructing a new ServerID()
+    // Other callers of new ServerID() are using getStringForm#getStringForm()
+    return (host.contains(":") ? ("[" + host + "]") : host) + DELIMITER + port;
```
myronkscott, line 91: shouldn't this be fixed where the IPv6 addresses are fed into this thing?  Seems like hostname should be a proper IPv6 address when this constructor is called

mathieucarbou: But it is already a proper address.

The problem is the concatenation with a port, it requires to be enclosed with square brackets.

myronkscott: when this is instantiated it should be `new Node("[::1]", 9410)` not `new Node("::1", 9410)`.  that makes more sense to me.

myronkscott: seems like a gray area.  [This](https://www.rfc-editor.org/rfc/rfc2732) seems to imply square brackets is valid for URLs but not the address itself for maybe we just fix this for network translator?

myronkscott: I'm really confused now because NetworkTranslator uses ServerID which should have the IPv6 address in the proper form. https://github.com/Terracotta-OSS/terracotta-core/blob/c23d4267ffe0ec58746eac628fc25ca66a5eef76/tc-server/src/main/java/com/tc/objectserver/impl/DistributedObjectServer.java#L1171

mathieucarbou: The ServerID ctor is called at 2 places. One is correctly formatting and the other place is constructing the ServerID from the method that I am fixing.

The code we did in DC is there because we ran into some bugs during the IPv6 system tests because of this concatenation

mathieucarbou: The spec says that square brackets are only required when using the IPv6 address concatenated to a port (I..e URI). Otherwise, this is not needed at all so the code in core is fine .

myronkscott: https://github.com/Terracotta-OSS/terracotta-core/blob/61dc22baf5677b4a10b6c1e43b81f5aee2e5b577/tc-server/src/main/java/com/tc/net/groups/TCGroupManagerImpl.java#L207-L222

This use of ServerID is only used in unit tests.  Can you point me to a failing test?  I want to investigate further.

mathieucarbou: I will have a look tomorrow. We have some IPv6 tests in DC in platform and EE and some passive redirection tests. And it's when we did some of those. It was more than a year ago now. Maybe core code has changed since...

mathieucarbou: So I've checked again and you're right the only place is in test code :  `new ServerID(new Node(hostname, port).getServerNodeName(), UUID.getUUID().toString().getBytes());`
I didn't pay attention.

But the good news then is that the problem we saw back then must be fixed and we can the remove the workaround and normally parse the string:-)

mathieucarbou:
Wrongly opened, the issue seems to be fixed now

PR #1299, opened 2022-11-16
simplify versioning

myronkscott:
Version strings can be defined elsewhere.  There is no format checking on version strings when created so core should not try to parse version strings.


tc-server/src/main/java/com/tc/objectserver/impl/ServerPersistenceVersionChecker.java
```diff
@@ -34,8 +33,8 @@ public class ServerPersistenceVersionChecker {
   }
 
   void checkAndBumpPersistedVersion(ClusterStatePersistor clusterStatePersistor) {
-    Version currentVersion = new Version(productInfo.version());
-    Version persistedVersion = clusterStatePersistor.getVersion();
+    String currentVersion = productInfo.version();
+    String persistedVersion = clusterStatePersistor.getVersion();
     // only move persisted version forward
     if (persistedVersion == null || currentVersion.compareTo(persistedVersion) > 0) {
```
tmesic99, line 39: `setVersion` in `ClusterStatePersister` used to set the version as `major.minor.micro`.  But now `setVersion` stores the entire string, including the 4 and 5th digits and anything else after that (e.g. -SNAPSHOT, -v0002, etc..).  
Is the `currentVersion.compareTo(persistedVersion) > 0` logic still valid given that before it only examined three digits but now it will evaluate against the entire string?  Not sure.

tmesic99: Maybe the `ServerPersistenceVersionCheckerTest` class should remain and modified to contain new examples of the full 5-digit version strings (plus optional suffix) that will be set in `setVersion`?

myronkscott: This class should be deleted

tmesic99: With the `ServerPersistenceVersionChecker` now removed, does the `VersionCompatibility` interface still need 'isCompatibleServerPersistence`?  Can that method be removed?

tmesic99 (approved):

PR #1305, opened 2023-01-18
Test fixes

myronkscott:


myronkscott:
Code cleanup.  Removing unused detach allows for deleting a bunch of broken code in BasicConnection and TCConnectoinImpl.  Added shutdown state to ElectionMgr because shutdown was doing more of a reset which was causing elections to re-run during shutdown causing servers to hang around longer than necessary.  Added code to stop spamming connection cleanup when the user opens a connection and discards it before closing the connection.

chrisdennis (approved):

PR #1308, opened 2023-01-23
[DAGGY] revert optimization that causes data corruption 

myronkscott:
due to improper use of byte buffers after recycle


chrisdennis (approved):
Is this related to the GC problem you hint at in the other PR? Could this be resurrected by ensuring the returned slice reference the existing buffer?

myronkscott:
I tried and it leads to too many OOME exhausting direct memory.  it's not GC'd quickly enough at least for tests.  It was a kind of pointless optimization anyways because it was eventually copied to byte[] because that is what the API calls for.

myronkscott:
I think we just merge this and move on.  If we want to try and optimize with less copying, that is a much larger effort

PR #1310, opened 2023-01-26
More data corruption

myronkscott:
This prevents data corruption when reading from the network by delaying buffer recycle until after hydrate of a TCAction has occurred.

### DIFF
--- a/common/src/main/java/com/tc/net/core/TCConnectionImpl.java
+++ b/common/src/main/java/com/tc/net/core/TCConnectionImpl.java
@@ -56,8 +56,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.LongAdder;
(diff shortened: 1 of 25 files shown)


myronkscott:
I updated only slightly to use close on the input stream rather than reset.  I started writing tests but the ones of consequence are taking too much mocking.  I tested this through EE locally and seems to work.  More tests and cleanup to come but I wanted to get this in now.

common/src/main/java/com/tc/net/core/TCConnectionImpl.java
```diff
@@ -211,9 +212,15 @@ private void closeImpl(Runnable callback) {
     }
     try {
       if (this.channel != null) {
-        this.commWorker.cleanupChannel(this.channel, callback);
+        CompletableFuture<Void> complete = new CompletableFuture<>();
+        this.commWorker.cleanupChannel(this.channel, ()->{
+          callback.run();
```
chrisdennis, line 217: Should probably be:
```java
try {
  callback.run();
  complete.complete(null);
} catch (Throwable t) {
  complete.completeExceptionally(t);
}
```

tc-messaging/src/main/java/com/tc/io/TCByteBufferInputStream.java
```diff
@@ -32,7 +32,7 @@ public class TCByteBufferInputStream extends InputStream implements TCByteBuffer
   private static final int            EOF                     = -1;
   private static final TCByteBuffer[] EMPTY_BYTE_BUFFER_ARRAY = new TCByteBuffer[] {};
 
-  private final Queue<TCByteBuffer>   returns;
+  private final Runnable   resetCommand;
```
chrisdennis, line 35: `closeCommand` or maybe `onCloseHook`?

tc-messaging/src/main/java/com/tc/io/TCByteBufferInputStream.java
```diff
@@ -316,11 +327,7 @@ private void nextBuffer() {
 
   @Override
   public void reset() {
```
chrisdennis, line 329: Can we remove this method (and the associated mark() method) from the interface now?

chrisdennis (changes requested):

tc-messaging/src/main/java/com/tc/io/TCByteBufferInputStream.java
myronkscott: yup

chrisdennis (approved):

PR #1312, opened 2023-02-13
DRAFT : Trying to grok reconnect interrupt handling

chrisdennis:
I think the reasserting here is wrong... and I think this is causing the occasional infinite reconnect loop. I wonder if the in to that loop is the line I've added a comment to where we interrupt and wait for termination, but don't flip any termination conditions. I'm thinking if this code path got called during a reconnect it could trigger one of these loops.

### DIFF
--- a/common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
+++ b/common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
@@ -294,7 +294,7 @@ public class ClientConnectionEstablisher {
       try {
         Thread.sleep(CONNECT_RETRY_INTERVAL);
       } catch (InterruptedException e1) {
-        Thread.currentThread().interrupt();
+        LOGGER.info("Async reconnect thread woken by interrupt " + transport);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
```diff
@@ -294,7 +294,7 @@ void handleConnectException(Exception e, boolean logFullException, Logger logger
       try {
         Thread.sleep(CONNECT_RETRY_INTERVAL);
       } catch (InterruptedException e1) {
-        Thread.currentThread().interrupt();
+        LOGGER.info("Async reconnect thread woken by interrupt " + transport);
       }
```
myronkscott, line 298: This seems problematic if a connection is closed while trying to connect or reconnect.  I'd like to understand why the connect loop is not breaking out of the loop if interrupted.  The only reason this thread should be interrupted is if the connection is closed.

chrisdennis: Right... but then what's going on with the interrupt that happens when `putConnectionRequest(...)` is called?

common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
```diff
@@ -442,6 +442,7 @@ private synchronized Thread getConnectionThread() {
     }
 
     public boolean putConnectionRequest(ConnectionRequest request) {
+      //TODO I'm worried about this code-path... it's signalling for a stop... but it's not setting any conditions.
```
myronkscott, line 445: not sure I understand this comment

PR #1313, opened 2023-02-15
update reconnect thread 

myronkscott:
Make attempts to reconnect cease if the thread  is interrupted.

Also did some cleanup and added a test.

### DIFF
--- a/common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
+++ b/common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
@@ -44,6 +44,7 @@ import java.util.Set;
 import java.util.function.Supplier;
 
 import static java.util.Arrays.asList;
+import java.util.Objects;
 
 /**
  * This guy establishes a connection to the server for the Client.
@@ -160,7 +161,7 @@ public class ClientConnectionEstablisher {
     return "ClientConnectionEstablisher[" + this.serverAddresses + "]";
   }
 
-  void reconnect(Supplier<Boolean> stopCheck) throws MaxConnectionsExceededException {
+  void reconnect(Supplier<Boolean> stopCheck) throws MaxConnectionsExceededException, InterruptedException {
     try {
       // Lossy logging for connection errors. Log the errors once in every 10 seconds
       LossyTCLogger connectionErrorLossyLogger = new LossyTCLogger(transport.getLogger(), 10000,
@@ -224,6 +225,11 @@ public class ClientConnectionEstablisher {
             }
             transport.reopen(target);
             connected = transport.isConnected() && transport.getConnectionID().isValid();
+            if (!connected) {
+              // if here, a strange state has been reached.  reopen should have either thrown an exception
+              // or created a proper connection.  Log it for now
+              LOGGER.warn("reopen succeeded however connection is not established ", new Exception());
+            }
           } catch (TransportRedirect redirect) {
             target = InetSocketAddress.createUnresolved(redirect.getHostname(), redirect.getPort());
           } catch (NoActiveException noactive) {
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
```diff
@@ -285,32 +291,28 @@ private boolean tryToConnect(boolean connected, Supplier<Boolean> stopCheck) {
     return !connected && !stopper && !stopped;
   }
 
-  void handleConnectException(Exception e, boolean logFullException, Logger logger) {
+  void handleConnectException(Exception e, boolean logFullException, Logger logger) throws InterruptedException {
     if (logger.isDebugEnabled() || logFullException) {
       logger.error("Connect Exception", e);
     }
 
     if (CONNECT_RETRY_INTERVAL > 0) {
```
chrisdennis, line 299: Logic in the constructor means this should never be < 1000. Think we can safely discard this if check.

common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
```diff
@@ -396,17 +406,19 @@ private class AsyncReconnect {
      * @param mayInterruptIfRunning interrupt thread
      * @return true if the passed in thread is joined
      */
-    private boolean waitForThread(Thread oldThread, boolean mayInterruptIfRunning) {
+    private boolean waitForConnectionThreadToFinish() {
```
chrisdennis, line 409: I think this method might be clearer structured like this:
```java
private boolean waitForConnectionThreadToFinish() {
  Thread oldThread = getConnectionThread();
  if (oldThread == null) {
    return true;
  } else if (Thread.currentThread() == oldThread) {
    LOGGER.warn("Assertion error, " + oldThread + " cannot join itself", new Exception());
    return false;
  } else {
    oldThread.interrupt();
    try {
       oldThread.join();
      return true;
    } catch (InterruptedException e) {
      LOGGER.info("Got interrupted while waiting for " + oldThread + " to complete");
      Thread.currentThread().interrupt();
      return false;
    }
  }
}
```

chrisdennis (changes requested):

myronkscott:
I added a commit to try and replace the thread creation with an executor and futures to clean things up a bit.  I think it is slightly more risky to make this change because of the runtime exceptions that executors can throw but i also think it makes things a bit clearer especially around double execution and startup and shutdown.  Let me know what you think.  I'm 50/50 if we should make this change.

common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
myronkscott: removed in the latest commit

common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
myronkscott: removed in the latest commit

myronkscott: this was poor logic because old thread often == current thread due to the fact that reconnects can be triggered by exceptions and disconnects during the reconnect process.

common/src/main/java/com/tc/net/protocol/transport/ClientConnectionEstablisher.java
```diff
@@ -372,111 +375,93 @@ private void quitReconnectAttempts() {
...
+        LOGGER.info("connect request ignored, reconnect has been shut down");
+        return false;
+      } else if (!currentTask.isDone()) {
+        //  if the current task is not done, this addtional request is likely 
+        //  due to a failed reconnect, let the original just continue on
+        LOGGER.info("connect request ignored, already reconnecting");
+        return false;
+      } else if (!transport.isConnected()) {
+        try {
+          currentTask = connectionEstablisher.submit(()->execute(request));
```
chrisdennis, line 442: Isn't this going to block because of the `SynchronousQueue` if two of these calls race?

myronkscott: no.  I used that queue as kind of a guard.  Since we are making sure the previous task is complete before adding a new one, the queue should be empty.

chrisdennis: Right but could there be two calls to this that both see `currentTask.isDone()` as `true and `transport.isConnected()` as `false`, and then they both end up submitting to the executor?

myronkscott: Trying to mimic what was there before.  Only one task ever in flight at one time.   The transport.isConnected check should be an assert really.  the code should never get to that point while being connected.

chrisdennis (approved):

PR #1314, opened 2023-02-16
Add reference counter for byte buffers implementation

myronkscott:


tc-messaging/src/main/java/com/tc/bytes/TCByteBufferAllocator.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Queue;
+import java.util.function.Supplier;
+
+/**
+ *
+ */
+public class TCByteBufferAllocator {
+  private final Supplier<TCByteBuffer> newBuffers;
+  private final Queue<TCByteBuffer> returns;
+  private final Collection<TCByteBuffer> items = new LinkedList<>();
```
chrisdennis, line 34: This obviously leans heavily on the ordering right, should probably type it as `List` just to be safe. Also why a `LinkedList`? `ArrayList` is probably a better choice, no?

tc-messaging/src/main/java/com/tc/bytes/TCReferenceSupport.java
```diff
@@ -0,0 +1,267 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ *
+ */
+public class TCReferenceSupport {
+  private final Queue<TCByteBuffer> returns;
```
chrisdennis, line 44: At this level this could just be a `Consumer<TCByteBuffer>` right?

tc-messaging/src/main/java/com/tc/bytes/TCReferenceSupport.java
```diff
@@ -0,0 +1,267 @@
...
+  private final Collection<TCByteBuffer> items;
+  private final AtomicInteger referenceCount = new AtomicInteger();
+  
+  // Ideally the below variables are not needed.  These are here to track references
+  // that are created but not closed, holding resources.
+  Optional<MemoryTracker> track;
```
chrisdennis, line 50: IntelliJ doesn't like this... I think it would prefer a nullable field.

tc-messaging/src/main/java/com/tc/bytes/TCReferenceSupport.java
```diff
@@ -0,0 +1,267 @@
...
+   * increases the reference count by one.  Every close decreases by one.  Once all the references
+   * are closed, the byte buffers are recycled back to the return queue.
+   * 
+   * @param tracked
+   * @param returns
+   * @return 
+   */
+  public static TCReference createReference(Collection<TCByteBuffer> tracked, Queue<TCByteBuffer> returns) {
+    if (returns == null) {
+      returns = new LinkedList<>();
```
chrisdennis, line 89: If we make this a `Consumer` then this would become `buffer -> {}` right?

tc-messaging/src/main/java/com/tc/bytes/TCReference.java
```diff
@@ -0,0 +1,32 @@
...
+ *  The Initial Developer of the Covered Software is
+ *  Terracotta, Inc., a Software AG company
+ *
+ */
+package com.tc.bytes;
+
+/**
+ *
+ */
+public interface TCReference extends Iterable<TCByteBuffer>, AutoCloseable {
```
chrisdennis, line 24: It's kinda bothering me that we're exposing both the array and the iterable... any way we could standardize on one of the other? I'm thinking that would simplify some of the implementations and prevent the need for conversion back and forth?

tc-messaging/src/main/java/com/tc/io/TCByteBufferInputStream.java
```diff
@@ -273,74 +164,68 @@ public final TCByteBuffer read(int len) {
...
 
-    while (this.index < this.numBufs) {
-      if (this.data[this.index].hasRemaining()) {
-        this.position++;
-        return this.data[this.index].get() & 0xFF;
+    if (len == 0) { 
+      return TCReferenceSupport.createGCReference(Collections.singleton(TCByteBufferFactory.getInstance(0))); 
+    }
+    
+    TCReference dup = data.duplicate();
```
chrisdennis, line 189: This code sitting outside of TCReference feels a bit wrong. Could we introduce say a `TCReference.duplicate(int length)` call that duplicates and truncates in a single call? That way we wouldn't have external code manipulating a member variable of the TCReference?

chrisdennis (changes requested):

tc-messaging/src/main/java/com/tc/bytes/TCByteBufferAllocator.java
myronkscott: ArrayList is better.

tc-messaging/src/main/java/com/tc/bytes/TCReferenceSupport.java
myronkscott: will change

tc-messaging/src/main/java/com/tc/bytes/TCReferenceSupport.java
myronkscott: was trying to avoid null checks but I guess I the code is doing it anyways, just in a different way

tc-messaging/src/main/java/com/tc/bytes/TCReferenceSupport.java
myronkscott: will change

tc-messaging/src/main/java/com/tc/bytes/TCReference.java
myronkscott: I'd like to get rid of the array stuff in a gradual way.  maybe depreciate asArray?

chrisdennis: Hmm... I was thinking of going the other way (maybe). The only thing tied heavily to the iterator right now is the read logic in the input stream that calls `iterator.remove()`. If you moved that inside the `TCReference` it might be easier to the array way and move to arrays for everything?

tc-messaging/src/main/java/com/tc/io/TCByteBufferInputStream.java
myronkscott: I can see that.  Feels weird manipulating the contents outside.  My justification for doing it this was to not to overload TCReference with too much functionality.  I wanted it to just be a container for ByteByffers.  Part of the problem is that a bunch of this stuff gets manipulated outside at all the layers.  TCByteBuffer.position and limit for example.  Even worse there is nothing preventing someone from holding on the the byte buffer and closing the reference.  Really, it would be nice to have just the reference and no bytebuffers but we are a long way from that.

tc-messaging/src/main/java/com/tc/bytes/TCReference.java
myronkscott: Actually, it looks like I culled most of the places asArray is used.  I'll try and remove it.

myronkscott: yeah.  It can be removed but I will do that in the adoption PR.

myronkscott: oh man.  Dealing with TCByteBuffer[] in the codebase is a total pain.  In a total contradiction to my earlier comment, I'd much rather have TCByteBuffer[] start migrating to TCReference.

myronkscott: I think it will be more clear when I push the adoption PR.

tc-messaging/src/main/java/com/tc/bytes/TCReferenceSupport.java
```diff
@@ -0,0 +1,273 @@
...
+ *
+ */
+public class TCReferenceSupport {
+  private final Consumer<TCByteBuffer> returns;
+  private final Collection<TCByteBuffer> items;
+  private final AtomicInteger referenceCount = new AtomicInteger();
+  
+  // Ideally the below variables are not needed.  These are here to track references
+  // that are created but not closed, holding resources.
+  MemoryTracker track;
```
chrisdennis, line 50: This can be final and private right?

chrisdennis (approved):

PR #1324, opened 2023-05-17
Stage config cleanup

myronkscott:
Cleanup some config code and add some additional logging.

### DIFF
--- a/common/src/main/java/com/tc/async/impl/StageManagerImpl.java
+++ b/common/src/main/java/com/tc/async/impl/StageManagerImpl.java
@@ -47,7 +47,8 @@ import java.util.concurrent.ConcurrentHashMap;
  * @author steve
  */
 public class StageManagerImpl implements StageManager {
-
+  private static final int         DEFAULT_MAX_STAGE_CAPACITY = TCPropertiesImpl.getProperties()
+                                                     .getInt(TCPropertiesConsts.L2_SEDA_STAGE_SINK_CAPACITY);
   private static final boolean     MONITOR       = TCPropertiesImpl.getProperties()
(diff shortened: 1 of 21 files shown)


common/src/main/java/com/tc/async/impl/StageManagerImpl.java
```diff
@@ -47,7 +47,8 @@
  * @author steve
  */
 public class StageManagerImpl implements StageManager {
-
+  private static final int         DEFAULT_MAX_STAGE = TCPropertiesImpl.getProperties()
```
prasantakumar82, line 50: If I understand correctly, this signifies capacity of sink of each SEDA stage i.e.  maximum size of queue. So why not naming it like `DEFAULT_STAGE_QUEUE_MAX_SIZE` or `DEFAULT_STAGE_SINK_CAPACITY`?

common/src/main/java/com/tc/net/core/TCConnectionImpl.java
```diff
@@ -419,19 +424,25 @@ private boolean buildWriteContextsFromMessages(boolean failfast) {
...
+          } catch (InterruptedException ie) {
+            Thread.currentThread().interrupt();
+            element = null;
+          }
         }
 
         if (MSG_GROUPING_ENABLED && batchMsgCount > 0) {
           final WireProtocolMessage ms = buildWireProtocolMessageGroup(currentBatch);
-          this.writeContexts.add(new WriteContext(ms));
+          this.writeContexts.add(new WriteContext(ms, batchMsgCount));
```
prasantakumar82, line 445: In continuation with my previous comment, will the instance of `WriteContext` contain number of messages in batch or total byte size of batch?

PR #1327, opened 2023-06-23
Remove unused thread local

myronkscott:


tc-server/src/main/java/com/tc/objectserver/entity/InvokeContextImpl.java
```diff
@@ -46,17 +46,6 @@ public InvokeContextImpl(ClientSourceId sourceId, int concurrencyKey, long oldes
     this.concurrencyKey = concurrencyKey;
     this.oldestid = oldestid;
     this.currentId = currentId;
-    setThreadLocal();
-  }
-
-  private void setThreadLocal() {
-    if (this.sourceId.isValidClient()) {
-      INHERITED.set(this);
```
chrisdennis, line 54: Any reason to not also delete the field itself?

chrisdennis (changes requested):

tc-server/src/main/java/com/tc/objectserver/entity/InvokeContextImpl.java
myronkscott: yeah.  oversight

chrisdennis (approved):

PR #1328, opened 2023-10-20
Remove callback port useage

myronkscott:
Removing handshake negotiated callback ports.  They don't work in heterogeneous networks such as when using containers.  Clients will still try and do socket connects as a last ditch effort to test server liveliness.  Servers will never try and socket connect clients.


mathieucarbou (approved):
Thanks!

PR #1331, opened 2024-03-21
Make the server more tolerant of client mistakes during reference/dereference

myronkscott:
The server crashes when it doesn't receive perfectly matched fetches and releases from the client.  While this will only happen due to a bug in the client accounting, Let's make the server more tolerant of these mistakes and push the failure to the client.  Also fixed improper accounting if the entity rejects reconnection.


tc-server/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
```diff
@@ -976,16 +976,28 @@ private void getEntity(ServerEntityRequest getEntityRequest, ResultCapture respo
...
       ClientDescriptorImpl descriptor = new ClientDescriptorImpl(clientID, getEntityRequest.getClientInstance());
       boolean added = clientEntityStateManager.addReference(descriptor, this.fetchID);
+      
+      if (canDelete) {
+        if (added) {
+          clientReferenceCount += 1;
+          Assert.assertTrue(clientReferenceCount > 0);
+        } else {
+          // this should never happen.  Ther server is expecting sane things from the client. 
+          logger.info("the client has attempted to fetch the same entity instance twice {}", descriptor);
```
chrisdennis, line 990: If it should never happen should it be a warn?

tc-server/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
```diff
@@ -1006,10 +1018,18 @@ private void getEntity(ServerEntityRequest getEntityRequest, ResultCapture respo
               handler.handleReconnect(descriptor, extendedData);
             }
           } catch (ReconnectRejectedException rejected) {
+            Assert.assertTrue(clientEntityStateManager.removeReference(descriptor));
```
chrisdennis, line 1021: Should the failure of this assertion (and others) be tunnelled to the client side too?

tc-client/src/main/java/com/tc/object/EntityClientEndpointImpl.java
```diff
@@ -220,9 +226,12 @@ public void didCloseUnexpectedly() {
     }
   }
 
-  private void checkEndpointOpen() {
+  private synchronized void checkEndpointOpen(boolean close) {
```
chrisdennis, line 229: Would seem simpler to keep `checkEndpointOpen()` as-is and have a new synchronized method for the close call?

tc-client/src/main/java/com/tc/object/CloseHookCallable.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package com.tc.object;
+
+import java.util.concurrent.Callable;
+import org.terracotta.exception.EntityException;
+
+/**
+ *
+ */
+public interface CloseHookCallable extends Callable<EntityException> {
```
chrisdennis, line 27: Do we need an interface here? What's wrong with just using the generic declaration directly?

chrisdennis (changes requested):

tc-server/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
myronkscott: I debated.  Was thinking of a certain consumer that considered new warn level logging an error in tests.

tc-server/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
myronkscott: This is a true assertion because the add happened inline.  The rejection is tunneled back to the client.

tc-client/src/main/java/com/tc/object/CloseHookCallable.java
myronkscott: The exception handling is messy.  Runnable doesn't really fit and I wanted a Callable that didn't throw.

myronkscott: We are just suppressing and/or logging these errors so I can move it to a generic callable

PR #1341, opened 2024-07-30
do not validate servers against standby nodes

myronkscott:


chrisdennis (approved):
This is a back port of a bug fix already in main, right?

PR #1343, opened 2024-08-06
[MAIN] Update Copyright

myronkscott:


tc-server/src/main/resources/com/tc/management/beans/TCServerInfoMBean.properties
```diff
@@ -1,21 +1,17 @@
...
-#  WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for
-#  the specific language governing rights and limitations under the License.
-#
-#  The Covered Software is Terracotta Core.
-#
-#  The Initial Developer of the Covered Software is
-#  Terracotta, Inc., a Software AG company
-#
-#
+ # Copyright Terracotta, Inc.
```
cljohnso, line 1: Does the Copyright item below need updating?

test-common/src/main/java/com/tc/util/Grep.java
```diff
@@ -1,22 +1,20 @@
 /*
+ *  Copyright Terracotta, Inc.
```
cljohnso, line 2: Should have a look at https://github.com/Terracotta-OSS/dso/blob/main/common/src/test/java/com/tc/util/Grep.java

tc-server/src/test/java/com/tc/net/protocol/transport/ServerStackProviderTest.java
```diff
@@ -1,19 +1,18 @@
 /*
+ *  Copyright Terracotta, Inc.
```
cljohnso, line 2: Stray copyright statement below this block ...

tc-server/src/test/java/com/tc/classloader/ServiceLocatorTest.java
```diff
@@ -1,19 +1,18 @@
 /*
+ *  Copyright Terracotta, Inc.
```
cljohnso, line 2: Stray copyright below this block

test-common/src/main/java/com/tc/util/Grep.java
myronkscott: going to just remove the class.  Not used

tc-server/src/main/resources/com/tc/management/beans/TCServerInfoMBean.properties
myronkscott: looks correct to me.  what do you see?

tc-server/src/test/java/com/tc/classloader/ServiceLocatorTest.java
myronkscott: will fix

tc-server/src/main/resources/com/tc/management/beans/TCServerInfoMBean.properties
cljohnso: There's a line later in the file against which I cannot make a review comment 

```java
Copyright            = Internal use only; Terracotta copyright information
```

Is this one correct?

myronkscott: Yes.  Correct.  That is the descriptor of the attribute.

cljohnso (approved):

PR #1355, opened 2024-12-11
Convert to gradle

myronkscott:


myronkscott:
Screenshot 2025-01-30 at 9 59 53 AM

This is what core will publish to maven.  client common and server APIs should be used with compileOnly and testImplementation.   Runtimes are published for packaging kits and testing.  server-runtime is now delivered as a zip archive of all the server jars in a flat directory.  galvan now embeds the server directly in its runtime and only requires plugin information to run.

Jenkinsfile
chrisdennis, line 1: If this is empty we should be able to delete it, right?

chrisdennis (approved):

PR #1381, opened 2025-09-15
Force Logback to INFO level 

aurbroszniowski:
due to default behaviour change in Logback 1.5 where DEBUG is now the new default


myronkscott (commented):
This looks good but don't we need to set the actual versions in the gradle.properties file?

aurbroszniowski:
> This looks good but don't we need to set the actual versions in the gradle.properties file?

Thanks for the comment, I changed the gradle properties file

tc-server/src/main/java/com/tc/l2/logging/TCLogbackLogging.java
```diff
@@ -122,6 +124,9 @@ public static void redirectLogging(File logDirFile) {
     } else {
       disableBufferingAppender(null);
     }
+    root.setLevel(Level.INFO);
```
myronkscott, line 127: Is there any way to know if this has been set via config file?  I would hate to always force INFO if someone specifically configured DEBUG or WARN

aurbroszniowski:
I updated the approach to keep the INFO level for the console and server logs.
Some DEBUG logs appeared before TCLogbackLogging is called, so the best solution I could find was to add a default logback.xml config in the classpath. Since Logback will pick up the first config that is found in the classpath, the user could still pass a custom config if he wants.

Also in
- TCLogbackLogging.resetLogging() added some code to make sure it reloads the xml config.
- TCLogbackLogging.redirectLogging() removed the temporary appender once full logging is configured, to avoid duplicate console output
- updated tests

I built a terracotta 12.0 enterprise kit including these terracotta-core changes and tested manually. The console and server are back to what they were with Logback 1.3, and I could pass my custom logback xml.

tc-server/src/main/java/com/tc/l2/logging/TCLogbackLogging.java
```diff
@@ -51,6 +51,12 @@ public static void resetLogging() {
     root.detachAndStopAllAppenders();
     console.detachAndStopAllAppenders();
     loggerContext.reset();
+
+    try {
+      new ch.qos.logback.classic.util.ContextInitializer(loggerContext).autoConfig();
+    } catch (Exception e) {
+      ch.qos.logback.core.util.StatusPrinter.printInCaseOfErrorsOrWarnings(loggerContext);
+    }
```
myronkscott, line 59: This is used for inline testing and is an attempt to shutdown everything for the current server/classloader.  The next server/classloader should do this automatically.

tc-server/src/main/java/com/tc/l2/logging/TCLogbackLogging.java
```diff
@@ -122,13 +128,19 @@ public static void redirectLogging(File logDirFile) {
     } else {
       disableBufferingAppender(null);
     }
+
+    Appender<ILoggingEvent> bootstrap = root.getAppender("TC_BASE");
+    if (bootstrap != null) {
+      root.detachAppender(bootstrap);
+      try { bootstrap.stop(); } catch (Exception ignore) {}
+    }
```
myronkscott, line 136: We still want console output going to the output stream provided to bootstrap logging.

PR #1383, opened 2025-10-07
logging extensions as plugins

myronkscott:
Add the ability to include plugins that configure Logback


galvan/src/main/java/org/terracotta/testing/master/ServerProcess.java
```diff
@@ -181,11 +181,8 @@ private synchronized void setStreams(OutputStream out, OutputStream err) {
 
   private String[] getJavaArguments(int debugPort) {
     // We want to bootstrap the variable with whatever is in our current environment.
-    String javaOpts = System.getenv("JAVA_OPTS");
     List<String> ops = new ArrayList<>();
-    if (javaOpts != null) {
-      ops.addAll(Arrays.asList(javaOpts.split(",")));
-    }
+
```
aurbroszniowski: just checking that removing the reading of the java opts is intentional?

aurbroszniowski (approved):

jhouserizer (approved):
Good cleanups and reorg

PR #1385, opened 2025-12-10
Refactor group manager and group config

myronkscott:
Removes some unused functionality.  

Creates a silent mode for discovery when there is no discovery on servers with a zero group port

update gradle pieces.


myronkscott:
This is some cleanup and a fix for rogue discovery during WAN testing

PR #1399, opened 2026-02-27
Bug fixes

myronkscott:
Fixes group manager hang.

Shutdown immediate needs to ignore stage and thread shutdown

Add PASSIVE-REPLICA-START mode to differentiate from PASSIVE-RELAY

### DIFF
--- a/tc-server/src/main/java/com/tc/net/groups/TCGroupManagerImpl.java
+++ b/tc-server/src/main/java/com/tc/net/groups/TCGroupManagerImpl.java
@@ -1258,9 +1258,12 @@ public class TCGroupManagerImpl implements GroupManager<AbstractGroupMessage>, C
           }
 
           @Override
-          public synchronized void complete() {
-            start();
-            signalComplete();
+          public void complete() {
+            try {
+              start();
+            } finally {
+              signalComplete();
+            }
           }
 
           private synchronized void signalComplete() {
--- a/tc-server/src/main/java/com/tc/l2/ha/L2HACoordinator.java
+++ b/tc-server/src/main/java/com/tc/l2/ha/L2HACoordinator.java
@@ -57,7 +57,7 @@ public class L2HACoordinator implements L2Coordinator, ConnectionIDFactoryListen
   private ReplicatedClusterStateManager                     rClusterStateMgr;
 
   public L2HACoordinator(Logger consoleLogger, DistributedObjectServer server,
-                         StateManager stateManager, 
+                         StateManager stateManager,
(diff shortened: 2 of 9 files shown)


mathieucarbou-ibm (approved):

mathieucarbou-ibm:
@myronkscott : is `PASSIVE-REPLICA-START` a state that can be observed from diagnostic port ? If yes, @guptakshit: do not forget to also add it to the LogicalState enum in platform. Thanks!

guptakshit:
> @myronkscott : is `PASSIVE-REPLICA-START` a state that can be observed from diagnostic port ? If yes, @guptakshit: do not forget to also add it to the LogicalState enum in platform. Thanks!

Yes, we’ll need to do that once these changes are released and the core version is upgraded in platform.

guptakshit (approved):

tc-server/src/main/java/com/tc/net/groups/TCGroupManagerImpl.java
```diff
@@ -1258,9 +1258,12 @@ private void start() {
           }
 
           @Override
-          public synchronized void complete() {
-            start();
-            signalComplete();
+          public void complete() {
+            try {
+              start();
```
devendra-ibm, line 1263: is it okay to complete the current state irrespective of any kind of exception. ?

devendra-ibm (approved):

tc-server/src/main/java/com/tc/net/groups/TCGroupManagerImpl.java
myronkscott: yes.   completion just indicates the next state can run.  This it true regardless of exception.

myronkscott:
> @myronkscott : is `PASSIVE-REPLICA-START` a state that can be observed from diagnostic port ? If yes, @guptakshit: do not forget to also add it to the LogicalState enum in platform. Thanks!

Yes.  This new state is observable.

tc-server/src/main/java/com/tc/net/groups/TCGroupManagerImpl.java
```diff
@@ -1258,9 +1258,12 @@ private void start() {
           }
 
           @Override
-          public synchronized void complete() {
-            start();
-            signalComplete();
+          public void complete() {
```
guptakshit, line 1261: will removing synchronized have any effect as it's present on method signalComplete() but not on method start()?

myronkscott: we only need the synchronized for notification.  There is only a single caller of complete.

PR #1400, opened 2026-03-02
Remove TCPropertyStore and back TCProperties with a JDK type

chrisdennis:


common/src/main/java/com/tc/properties/TCPropertiesImpl.java
```diff
@@ -62,30 +63,23 @@ public class TCPropertiesImpl implements TCProperties {
...
-  private final TCPropertyStore         props                      = new TCPropertyStore();
+  private final Map<String, String>         props                      = new LinkedHashMap<>();
 
-  private static final ThreadLocal<TCPropertyStore>         localTcProperties          = new InheritableThreadLocal<>() {
-    @Override
-    protected TCPropertyStore initialValue() {
-      return new TCPropertyStore();
-    }
-  };
+  private final ThreadLocal<Map<String, String>>         localTcProperties          = new InheritableThreadLocal<>();
```
myronkscott, line 68: this shouldn't be static?

myronkscott (approved):
Just one question.  Modernization is good.  I'm tempted to create a TCThreadGroup local and store the values there.

common/src/main/java/com/tc/properties/TCPropertiesImpl.java
chrisdennis: This becomes static via the singleton nature of TCProperties. I assumed that if we ever want more TCPropertiesImpl instances they would need independent thread local properties.

jhouserizer (approved):
looks safe to me!

PR #1402, opened 2026-03-07
Fix shutdown

myronkscott:
The purpose of this PR is to reimplement and reorganize shutdown code to be more reliable under both process and inline servers.  

Inline servers must always observe full organized shutdown to work properly while process servers can sometimes require immediate shutdown where the server just exits.  

Here we pull all the extra thread waiting and shutdown code into TCServer from DSO more clearly separating responsibility.  

DSO "shutdown" via state change which triggers shutdown of each subunit.

TCServerImpl is responsible for waiting until; DSO is complete then making sure all the threads of TCThreadGroup are retired and shutdown if immediate shutdown is not asked for or if inline server running.


common/src/main/java/com/tc/properties/TCPropertiesImpl.java
```diff
@@ -149,20 +154,27 @@ private void applyConfigOverrides(Map<String, String> overwriteProps) {
...
+                      + props.get(key) + ". This will not be overridden to " + propertyValue
+                      + " from the tc-config file");
         }
-        locals.put(key, value);
       } else {
-        logger.warn("The property \"" + propertyName + "\" was set by local settings to "
-                    + props.get(key) + ". This will not be overridden to " + propertyValue
-                    + " from the tc-config file");
+        props.put(key, value);
+        logger.info("The property \"" + propertyName + "\" was overridden to " + propertyValue + " from "
```
jhouserizer: Better wording!

jhouserizer (approved):
Refactoring of thread/logic all looks ok to me.

chrisdennis (changes requested):
When doing a gradle upgrade you need to run the wrapper task twice... or more specifically:
```bash
$ ./gradlew wrapper --gradle-version=8.14.4
$ ./gradlew wrapper
```
The first invocation just updates the properties file... the second one will fetch the new jars and bash/windows scripts.

myronkscott:
> When doing a gradle upgrade you need to run the wrapper task twice... or more specifically:
> 
> ```shell
> $ ./gradlew wrapper --gradle-version=8.14.4
> $ ./gradlew wrapper
> ```
> 
> The first invocation just updates the properties file... the second one will fetch the new jars and bash/windows scripts.



> When doing a gradle upgrade you need to run the wrapper task twice... or more specifically:
> 
> ```shell
> $ ./gradlew wrapper --gradle-version=8.14.4
> $ ./gradlew wrapper
> ```
> 
> The first invocation just updates the properties file... the second one will fetch the new jars and bash/windows scripts.

Thanks.  Fixed

chrisdennis (approved):

mathieucarbou (approved):

PR #1404, opened 2026-03-25
TDB-19412: Fix properties object returned without the incoming properties when cif is null

mathieucarbou-ibm:
This is a proposed fix for https://github.com/Terracotta-OSS/terracotta-platform/pull/1268#issuecomment-4127830174 but I have no clue how much it can impact other places


jhouserizer (approved):
would be nice to have Myron comment on whether any unexpected side-effect could come in other scenarios

myronkscott (approved):

PR #1406, opened 2026-04-06
Invoke Context for server messages

myronkscott:
Server messages should inherit its invoke context from the calling context.  

Replica invokes should not use client's as they will never reconnect.

### DIFF
--- a/tc-server/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
+++ b/tc-server/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
@@ -1,6 +1,6 @@
 /*
  *  Copyright Terracotta, Inc.
- *  Copyright IBM Corp. 2024, 2025
+ *  Copyright IBM Corp. 2024, 2026
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
@@ -28,6 +28,7 @@ import com.tc.exception.TCServerRestartException;
 import com.tc.exception.TCShutdownServerException;
 import com.tc.l2.msg.SyncReplicationActivity;
 import com.tc.net.ClientID;
+import com.tc.net.protocol.tcm.MessageChannel;
 import com.tc.net.utils.L2Utils;
 import com.tc.object.ClientInstanceID;
 import com.tc.object.EntityDescriptor;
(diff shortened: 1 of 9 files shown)


tc-server/src/main/java/com/tc/objectserver/handler/ReplicatedTransactionHandler.java
```diff
@@ -540,14 +543,14 @@ private void moveToPassiveStandBy() {
 
   private ServerEntityRequest activityToLocalRequest(SyncReplicationActivity activity) {
     ActivityType activityType = activity.getActivityType();
-    ClientID source = activity.getSource();
-    ClientInstanceID instance = activity.getClientInstanceID();
+    ClientID source = isReplica && activityType == ActivityType.INVOKE_ACTION ? ClientID.NULL_ID : activity.getSource();
```
devendra-ibm, line 546: I see, this fixes the original issue, and it's protecting original clientID for lifecycle messages, as lifecycle messages don't get tracked any way, but I didn't understand the changes related to server-derived messages, i see previously these messages didn't inherit the original request clientIds or txnIds, but now it's preserving it. why it's needed now.?

myronkscott: Because store creates a synthetic InvokeContext when invoking server initiated messages on the passive.  We need to abandon that and essentially do the same thing in the server rather than the entity so that we can change the value when the server is operating as a replica.

devendra-ibm: okay, make sense. then we might need to remove synthetic InvokeContext construction from DatasetPassiveEntity as well,  since, now we're sending real InvokeContext so it will override the InvokeContext associated values for replica as well, i.e. it's going to replace NULL clientID with original clientId for replica.

tc-server/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
```diff
@@ -466,12 +469,45 @@ private EntityMessage decodeMessage(MessagePayload payload, ResultCapture captur
...
+
+    if (local == null) {
+      local = (ThreadLocal<ServerEntityRequest>) channel.getAttachment(REQUEST_CONTEXT_KEY);
+      if (local == null) {
+        local = new ThreadLocal<>();
+        channel.addAttachment(REQUEST_CONTEXT_KEY, local, false);
+      }
+    }
+
+    local.set(request);
```
guptakshit, line 485: do we require double null check here?

myronkscott: This could be better.  I'll clean it up

devendra-ibm (approved):
we need this PR for DR testing.

devendra-ibm:
apart this PR gets merged, we might need to remove synthetic invokeContex from ClusterTierPassiveEntity and datasetPassiveEntity as well, as we have see similar data loss in ehcache as well.

guptakshit (approved):

PR #1413, opened 2026-05-08
Fixes for zapping replica when out of sync

myronkscott:
This PR fixes replica servers when they are out of sync.  These servers need to be Zapped (dirtydb) rather than simply restarted when it is out if sync with the relay.

Also fixes an NPE error in certain error conditions.

### DIFF
--- a/tc-server/src/main/java/com/tc/l2/ha/L2HAZapNodeRequestProcessor.java
+++ b/tc-server/src/main/java/com/tc/l2/ha/L2HAZapNodeRequestProcessor.java
@@ -1,6 +1,6 @@
 /*
  *  Copyright Terracotta, Inc.
- *  Copyright IBM Corp. 2024, 2025
+ *  Copyright IBM Corp. 2024, 2026
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


tc-server/src/main/java/com/tc/l2/ha/L2HAZapNodeRequestProcessor.java
```diff
@@ -162,7 +162,6 @@ public void incomingZapNodeRequest(NodeID nodeID, int zapNodeType, String reason
         String message = "Terminating due to Zap request from " + getFormatedError(nodeID, zapNodeType, reason);
         logger.error(message);
         if (zapNodeType == NODE_JOINED_WITH_DIRTY_DB) {
-          clusterStatePersistor.setDBClean(false);
```
devendra-ibm, line 165: I didn't understand this, why we don't need to persist db clean state now or we persisting from somewhere else .?

myronkscott: The zap exception handler does this.  Just cleaning up setting this in two places.

devendra-ibm (approved):
approved.

guptakshit (approved):

PR #1416, opened 2026-05-18
Add dependency constraints for external dependencies to server-runtime

chrisdennis:
This should allow us to use the server-runtime as a platform/BOM in the Gradle builds and get enforcement of its external transitive dependencies, without changing the way things are delivered. I still need to do some testing downstream to make sure things work - but I think this will solve my problems.

### DIFF
--- a/server-runtime/build.gradle
+++ b/server-runtime/build.gradle
@@ -1,7 +1,9 @@
+import org.gradle.api.internal.project.ProjectInternal
+import org.gradle.api.plugins.jvm.internal.JvmPluginServices
 
 /*
  * Copyright Terracotta, Inc.
- * Copyright IBM Corp. 2024, 2025
+ * Copyright IBM Corp. 2024, 2026
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


server-runtime/build.gradle
```diff
@@ -16,33 +18,50 @@
...
+def serverBucket = configurations.dependencyScope("server")
+def serverClasspath = configurations.resolvable("serverClasspath") {
+    extendsFrom(serverBucket.get())
+    ((ProjectInternal) project).services.get(JvmPluginServices.class).configureAsRuntimeClasspath(it)
+}
+
 description = 'terracotta-core server-runtime'
 
 dependencies {
-  compileOnlyApi project(':server-api')
```
myronkscott, line 27: The server-api is not published as part of the api in server-runtime-\<VERSION\>.module.  Do we not care about that?  Are we saying that server-api will be used directly by consuming plugins?  If we include it, can we use it when making a kit to validate compatibility?

myronkscott (approved):

server-runtime/build.gradle
chrisdennis: This declaration wasn't actually doing anything in the first place. The server-api dependency wasn't ending up in any of the metadata, pom or module. We could do more rework in this area to do smart things with apiElements vs runtimeElements, but it might make things tricky with the zip packaging we're using for runtime.

chrisdennis:
In kits now this allows us to do:
```kotlin
dependencies {
  serverRuntime(enforcedPlatform(terracotta.core.server))
  serverRuntime(terracotta.core.server)
}
```
This enforces that the version of the external dependencies (currently slf4j-api and logback) of the server runtime must be used by the kit.

PR #1422, opened 2026-06-04
save a sha256 with the unpack of galvan-test-server 

myronkscott:
check the checksum each time the deploy is called to make sure the files are current

### DIFF
--- a/galvan/build.gradle
+++ b/galvan/build.gradle
@@ -51,30 +51,30 @@ def galvanKit = tasks.register('galvanServerZip', Zip) {
 }
 
 def generateChecksum = tasks.register('generateGalvanChecksum') {
-    dependsOn galvanKit
-    
-    def checksumFile = layout.buildDirectory.file('galvan-kit/galvan-test-server.zip.sha256')
-    
-    inputs.files(galvanKit.map { it.archiveFile })
-    outputs.file(checksumFile)
+    def inputFile = galvanKit.flatMap { it.archiveFile}
+    def checksumDir = layout.buildDirectory.dir("galvan-checksums")
+
+    inputs.file(inputFile).withPathSensitivity(PathSensitivity.NONE)
+    outputs.dir(checksumDir)
     
     doLast {
-        def zipFile = galvanKit.get().archiveFile.get().asFile
+        def zipFile = inputFile.get().asFile
+        def checksumFile = checksumDir.get().file("${zipFile.name}.sha256").asFile
         def checksum = java.security.MessageDigest.getInstance('SHA-256')
-            .digest(zipFile.bytes)
-            .encodeHex()
-            .toString()
-        checksumFile.get().asFile.text = checksum
+
+        zipFile.eachByte(8192) { byte[] buffer, int length ->
+            checksum.update(buffer, 0, length)
+        }
+
+        checksumFile.withWriter {
+            checksum.digest().encodeHex().writeTo(it)
+        }
     }
 }
 
 sourceSets.main {
-    resources.srcDir(galvanKit.map { it.destinationDirectory })
-}
-
-// Ensure checksum is generated before processing resources
-tasks.named('processResources') {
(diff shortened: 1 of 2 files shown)


mathieucarbou-ibm (approved):

galvan/build.gradle
```diff
@@ -50,10 +50,33 @@ def galvanKit = tasks.register('galvanServerZip', Zip) {
     }
 }
 
+def generateChecksum = tasks.register('generateGalvanChecksum') {
```
chrisdennis, line 53: Better Gradle here would be:
```groovy
def generateChecksum = tasks.register('generateGalvanChecksum') {
    def inputFile = galvanKit.flatMap { it.archiveFile}
    def checksumDir = layout.buildDirectory.dir("galvan-checksums")

    inputs.file(inputFile).withPathSensitivity(PathSensitivity.NONE)
    outputs.dir(checksumDir)
    
    doLast {
        def zipFile = inputFile.get().asFile
        def checksumFile = checksumDir.get().file("${zipFile.name}.sha256").asFile
        def checksum = MessageDigest.getInstance('SHA-256')

        zipFile.eachByte(8192) { byte[] buffer, int length ->
            checksum.digest(buffer, 0, length)
        }

        checksumFile.withWriter {
            checksum.digest().encodeHex().writeTo(it)
        }
    }
}

sourceSets.main {
    output.dir(galvanKit)
    output.dir(generateChecksum)
}
```
and no need to have `processResources` depend on `generateChecksum`.

galvan/src/main/java/org/terracotta/testing/master/ServerDeploymentBuilder.java
```diff
@@ -111,18 +114,27 @@ public Path deploy(boolean refresh) {
     try (FileChannel dc = FileChannel.open(dl, StandardOpenOption.CREATE, StandardOpenOption.WRITE)) {
       while (true) {
         try (FileLock lock = dc.lock()) {
-          if (Files.find(ip, 10, (path,attr)->path.getFileName().toString().startsWith("tc-server")).findAny().isPresent()) {
-            if (refresh) {
+          String currentChecksum = calculateZipChecksum();
+          String storedChecksum = readStoredChecksum(ip);
+
+          boolean needsRefresh = refresh ||
+              !Files.find(ip, 10, (path,attr)->path.getFileName().toString().startsWith("tc-server")).findAny().isPresent() ||
```
chrisdennis: Do we need to do this check here? Won't the checksum comparison do this for you?

chrisdennis (changes requested):

galvan/src/main/java/org/terracotta/testing/master/ServerDeploymentBuilder.java
myronkscott: interesting question.  I guess if one is there the other will be too.

galvan/build.gradle
myronkscott: curious why Galvan-checksums dir is better

chrisdennis: It's best practice to have each task write to a separate output directory to avoid confusion with the up-to-date-ness of task inputs and outputs.

myronkscott: I got it, puts it into a dir for the sourceSets to pickup.

chrisdennis (approved):

galvan/build.gradle
```diff
@@ -50,8 +50,31 @@ def galvanKit = tasks.register('galvanServerZip', Zip) {
...
+
+        checksumFile.withWriter {
+            checksum.digest().encodeHex().writeTo(it)
+        }
+    }
+}
+
 sourceSets.main {
-    resources.srcDir(galvanKit.map { it.destinationDirectory })
+    output.dir(galvanKit)
```
myronkscott: @chrisdennis This change doesn't work for tests.  I can't claim to understand why.

PR #1423, opened 2026-06-15
Fix invoke context for client requests

myronkscott:
Invoke context needs refactoring to include clientID and transaction id to track where messages originate.  We also need to refactor invoke responses to route to the appropriate captures based on whether the request came from a client or is server generated.  Also cleanup up code so multiple functions are clearly separated into different classes.  Request, Response, ResultCapture and Retiree are all separated rather than combined into a superclass handling all functions.


myronkscott:
Also fixed exporting as API a package that is internally defined and not published

tc-messaging/src/main/java/com/tc/entity/NetworkVoltronEntityMessageImpl.java
chrisdennis, line 1: I assume you modified this at one point... but it looks like this is all just whitespace changes now.

tc-messaging/src/main/java/com/tc/entity/ResendVoltronEntityMessage.java
chrisdennis, line 1: Looks to be just whitespace changes again.

tc-messaging/src/main/java/com/tc/entity/VoltronEntityMessage.java
```diff
@@ -90,36 +92,52 @@ enum Acks {
...
+  /**
+   * Server originated messages need to be handled differently from client initiated
+   * @return true if the message is from a server.
+   *
+   */
+  default boolean isServerRequest() {
+    return false;
+  }
+
+  default Consumer<byte[]> getCompletionHandler() {
```
chrisdennis, line 136: Would it be cleaner if these were:
```
return value -> {}
```

tc-server/src/main/java/com/tc/objectserver/api/ResultCapture.java
```diff
@@ -27,6 +30,73 @@
...
   void message(byte[] message);
   void setWaitFor(Supplier<ActivePassiveAckWaiter> waiter);
-  void waitForReceived();
   CompletionStage<Void> retired();
+
+  static ResultCapture chain(ResultCapture...list) {
+    return new ResultCapture() {
+      @Override
+      public void message(byte[] message) {
+        for (ResultCapture r : list) {
```
chrisdennis, line 38: It looks to me like everything passed here is non-null for the moment. Can we not just remove these null-checks and assume that?

tc-server/src/main/java/com/tc/objectserver/api/ResultCapture.java
```diff
@@ -27,6 +30,73 @@
...
+          if (r != null) {
+            r.setWaitFor(waiter);
+          }
+        }
+      }
+
+      @Override
+      public CompletionStage<Void> retired() {
+        CompletionStage<Void> together = CompletableFuture.completedFuture(null);
```
chrisdennis: Rather than chaining this off a synthetic future, could we just do a stream reduction?

```java
return Arrays.stream(list).map(ResultCapture::retired)
        .reduce((a, b) -> a.thenCombine(b, (x, y) -> null))
        .orElse(CompletableFuture.completedFuture(null));
```

tc-server/src/main/java/com/tc/objectserver/api/ServerEntityRequest.java
chrisdennis, line 1: Whitespace only.

tc-server/src/main/java/com/tc/objectserver/entity/LifecycleResultCapture.java
```diff
@@ -0,0 +1,162 @@
...
+        ActivePassiveAckWaiter waiter = setOnce.get();
+        waiter.waitForCompleted();
+        if (waiter.verifyLifecycleResult(false)) {
+          LOGGER.warn("ZAP occurred while processing " + request.getAction() + " on " + this.descriptor);
+        }
+      }
+    }
+
+    @Override
+    public void complete() {
```
chrisdennis, line 116: My brain wants to flip these two around so thats it's clear that we never use `value`.

tc-server/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
```diff
@@ -906,9 +907,14 @@ private void performAction(ServerEntityRequest wrappedRequest,
...
       } else {
         this.retirementManager.registerWithMessage(message, concurrencyKey, new Retiree() {
+          private final SetOnceFlag retired = new SetOnceFlag();
           @Override
           public CompletionStage<Void> retired() {
-            return response.retired();
+            if (retired.attemptSet()) {
+              return response.retired();
+            } else {
+              throw new IllegalStateException("already retired");
```
chrisdennis, line 916: Why wouldn't this just return a completed future?

tc-server/src/main/java/com/tc/objectserver/entity/NetworkInvokeResponse.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    @Override
+    public void complete(byte[] result) {
+      stats[StatType.SERVER_COMPLETE.serverSpot()] = System.nanoTime();
+      sendResponse(result);
+    }
+
+    @Override
+    public void complete() {
```
chrisdennis, line 112: Should this just be `complete(new byte[0])`?

tc-server/src/main/java/com/tc/objectserver/entity/NoopResultCapture.java
```diff
@@ -19,64 +19,55 @@
...
-  public void waitForReceived() {
+  public void setWaitFor(Supplier<ActivePassiveAckWaiter> waiter) {
 
   }
 
   @Override
-  public void received() {
-    Trace.activeTrace().log("Received");
+  public CompletionStage<Void> retired() {
+    return CompletableFuture.completedFuture(null);
```
chrisdennis: Does it make sense for this to be a final instance variable so we're not creating a new instance every time?

tc-server/src/main/java/com/tc/objectserver/entity/ReconfigureMessage.java
chrisdennis, line 1: Copyright only.

tc-server/src/main/java/com/tc/objectserver/entity/ServerEntityRequestImpl.java
chrisdennis, line 1: Whitespace only.

tc-server/src/test/java/com/tc/objectserver/entity/EntityManagerImplTest.java
chrisdennis, line 1: Whitespace only.

tc-server/src/test/java/com/tc/objectserver/handler/ProcessTransactionHandlerTest.java
```diff
@@ -325,6 +325,11 @@ private NetworkVoltronEntityMessage createMockRequestWithSender(VoltronEntityMes
     when(request.getEntityDescriptor()).thenReturn(entityDescriptor);
     when(request.getTransactionID()).thenReturn(transactionID);
     when(request.getOldestTransactionOnClient()).thenReturn(new TransactionID(1));
+    if (sender.isNull()) {
```
chrisdennis:
```java
when(request.isServerRequest()).thenReturn(sender.isNull())
```

tc-server/src/test/java/com/tc/objectserver/handler/RetirementManagerTest.java
chrisdennis, line 1: Whitespace only.

chrisdennis (changes requested):

tc-server/src/main/java/com/tc/objectserver/entity/ManagedEntityImpl.java
myronkscott: calling retire on this twice is a bug.

tc-server/src/test/java/com/tc/objectserver/handler/ProcessTransactionHandlerTest.java
myronkscott: Mockito doesn't like the dynamic nature of checking isNull() in the return

chrisdennis (approved):
Approved, but two things to note:
1. Handful of classes still just seem to contain reformatting changes. Not wrong but not sure if it's intended or not.
2. `isServerRequest()` is only ever used in a negated form `!isServerRequested()` - do we want to consider flipping the sense and renaming it?

myronkscott:
renamed the method as suggested.  The remaining whitespace and formatting changes are due to old commits that changed the file but missed whitespace checks.

PR #1424, opened 2026-06-17
Invoke messenger to replace IEntityMessenger

myronkscott:
There are several problems with IEntityMessenger.  

1.  Context of the originating message is not preserved.  We can't track the message sent back to the originating invoke message.
2. It's technically possible to send server messages from a passive entity which should not be allowed.

Here we replace the IEntityMessenger service with the ability to create ActiveServerMessenger which is able to preserve the originating message's context such as which client/transaction the message belongs to.  Automatic deferment of the originating message until all the messages sent under the originating context complete.


myronkscott:
@chrisdennis This is a draft PR.  Please comment on the changes under terracotta-apis only.  Thanks!

myronkscott:
Pruned back this change to just move the defer retirement messages so they run in the context for the owning parent message.  Retirement can now only be deferred for the parent message in the context.  There is no longer an opportunity to defer retirement of server initiated messages.  Server initiated messages can still be sent but they will have no parent message or owning context.

myronkscott:
Using Consumers for the sendMessage results so they re delivered by the thread that ran the command.  Don't want the calling thread blocking with a future.

terracotta-apis/entity-server-api/src/main/java/org/terracotta/entity/ActiveServerMessenger.java
```diff
@@ -0,0 +1,88 @@
...
+  void sendMessage(EntityMessage message, Consumer<Response<R>> result);
+  /**
+   * Defer retirement of the current context message using the message as the trigger.
+   *
+   * @param tag debugging tag for the handle
+   * @param message release trigger message.  It is not scheduled until the release handle
+   * is released
+   * @return a release handle to release retirement of the current message
+   */
+  ReleaseHandle deferRetirement(String tag, EntityMessage message);
```
chrisdennis, line 46: Do we still need these two methods if we have `sendMessage(...)`? Could we eliminate them?

terracotta-apis/entity-server-api/src/main/java/org/terracotta/entity/ActiveServerMessenger.java
```diff
@@ -0,0 +1,88 @@
...
+   * after the release handle is released
+   * @return a release handle to release retirement of the current message
+   */
+  ReleaseHandle deferRetirement(String tag, EntityMessage message, Consumer<Response<R>> result);
+
+  /**
+   * {@link AutoCloseable} without exception.
+   */
+  @Override
+  void close();
```
chrisdennis, line 62: Implementations don't seem to use/need this close method.

terracotta-apis/standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
```diff
@@ -48,62 +48,29 @@ public interface IEntityMessenger<M extends EntityMessage, R extends EntityRespo
...
-   * Asynchronously send a message to the entity instance which looked up the service instance.
-   * 
-   * @param message The message to send.
-   * @param response A callback used when the result of the invoke is available.  NOTE: callback delivered
-   * on the thread which the invoke occurs
-   * @throws MessageCodecException The message could not be serialized.
-   */  
-  void messageSelf(M message, Consumer<MessageResponse<R>> response) throws MessageCodecException;
-
+  void messageSelf(M message);
```
chrisdennis, line 56: We're moving MessageCodecException to `UncheckedIOException` right? If we're not going to float it up through the API do we want to consider eliminating it entirely?

chrisdennis (changes requested):
API review/thoughts. There a bunch more fine detail stuff in the implementation, but I figure that can wait until the PR is out of draft - most of it is minor/cosmetic.

terracotta-apis/entity-server-api/src/main/java/org/terracotta/entity/ActiveServerMessenger.java
myronkscott: These are slightly different.  These allow you to queue a message and defer the current invoke.  The message is sent when the handle is released.  I think we still need it.

terracotta-apis/standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
myronkscott: I think it is still useful and a fatal exception as if your installation is broken and the server should be shutdown.

terracotta-apis/entity-server-api/src/main/java/org/terracotta/entity/ActiveServerMessenger.java
myronkscott: I probably should add implementation that throws or somehow fails to send on a closed connection.  I think this is important for scope because we don't want someone holding on to an ActiveInvokeMessenger across invoke calls out of the scope of the current message invoke.

terracotta-apis/standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
chrisdennis: I'm mainly thinking if the MessageCodecException isn't ever thrown to a user... then do we need the exception to exist at all? Can we not just throw anything we like out of the codec code?

terracotta-apis/entity-server-api/src/main/java/org/terracotta/entity/ActiveServerMessenger.java
chrisdennis: You mean once the originating message is retired right? If the message is retired then all the other methods should fail

galvan-support/src/test/java/org/terracotta/functional/ActiveMessengerFunctionIT.java
```diff
@@ -0,0 +1,52 @@
...
+  public void testActiveMessenger() throws Exception {
+    Connection connection = CLUSTER.newConnection();
+    String hp = CLUSTER.getClusterHostPorts()[0];
+    String[] shp = hp.split(":");
+    int port = Integer.parseInt(shp[1]);
+
+    EntityRef<ConcurrentClusteredMap, MapConfig, Void>  ref = connection.getEntityRef(ConcurrentClusteredMap.class, 1L, "ROOT");
+    ref.create(new MapConfig(1, "ROOT"));
+    ConcurrentClusteredMap map = ref.fetchEntity(null);
+    map.putMultiple(new HashMap<>());
```
chrisdennis, line 50: Should we be validating the passive contents here?

galvan-support/src/test/java/org/terracotta/functional/ActiveMessengerFunctionIT.java
```diff
@@ -0,0 +1,52 @@
+/*
+ *  Copyright Terracotta, Inc.
+ *  Copyright IBM Corp. 2024, 2026
```
chrisdennis, line 3: Just 2026?

passthrough-server/src/main/java/org/terracotta/passthrough/PassThroughServerActiveInvokeContext.java
```diff
@@ -16,33 +16,41 @@
...
+  private final String entityName;
 
-  public PassThroughServerActiveInvokeContext(M message, ClientDescriptor descriptor, int concurrencyKey, long current, long
-    oldest, IMessageSenderWrapper monitor, PassthroughRetirementManager retirement, MessageCodec<M, R> codec) {
+  public PassThroughServerActiveInvokeContext(M message, PassthroughClientDescriptor descriptor, int concurrencyKey, long current, long
+    oldest, IMessageSenderWrapper monitor, PassthroughRetirementManager retirement, MessageCodec<M, R> codec, PassthroughServerProcess process, String entityClass, String entityName) {
     super(descriptor == null ? null : descriptor.getSourceId(), concurrencyKey, current, oldest);
-    this.message = message;
-    this.descriptor = descriptor;
+    this.parent = Objects.requireNonNull(message);
```
chrisdennis, line 46: This looks to be unused now.

passthrough-server/src/main/java/org/terracotta/passthrough/PassthroughMessengerService.java
```diff
@@ -21,11 +21,10 @@
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.io.UncheckedIOException;
```
chrisdennis, line 24: Unused.

passthrough-server/src/main/java/org/terracotta/passthrough/PassthroughMessengerService.java
chrisdennis, line 37: Unused?

passthrough-server/src/main/java/org/terracotta/passthrough/PassthroughMessengerService.java
```diff
@@ -39,7 +38,7 @@ public class PassthroughMessengerService implements IEntityMessenger<EntityMessa
   private final DeferredEntityContainer entityContainer;
   private final String entityClassName;
   private final String entityName;
-    
+
   public PassthroughMessengerService(PassthroughTimerThread timerThread, PassthroughServerProcess passthroughServerProcess, DeferredEntityContainer entityContainer, boolean chain, String entityClassName, String entityName) {
```
chrisdennis, line 42: `chain` seems to be unused which can then cleanup the call-site too.

tc-server/src/main/java/com/tc/objectserver/entity/ActiveInvokeContextImpl.java
```diff
@@ -18,34 +18,41 @@
...
+import org.slf4j.LoggerFactory;
 import org.terracotta.entity.ActiveInvokeChannel;
 import org.terracotta.entity.ActiveInvokeContext;
 import org.terracotta.entity.ClientDescriptor;
+import org.terracotta.entity.EntityMessage;
 import org.terracotta.entity.EntityResponse;
+import org.terracotta.entity.ActiveServerMessenger;
 
 public class ActiveInvokeContextImpl<R extends EntityResponse> extends InvokeContextImpl implements ActiveInvokeContext<R> {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ActiveInvokeContextImpl.class);
```
chrisdennis, line 38: Unused.

chrisdennis (changes requested):
Handful of dead code. Otherwise this looks good. There's a bunch of raw typing going on here too... I don't think any of it is hiding any real issues though. Do we ever want to clean that up or are we good with it staying around?

chrisdennis (approved):

jhouserizer (approved):
I at least skimmed through all of this, looked at quite a lot of it (changes to exception throwing, etc.), and spent a fair amount of time reviewing the couple of areas that I thought most sensitive (such as msg retirement).

PR #1430, opened 2026-07-30
ThreadGroup destroy is deprecated.

myronkscott:
remove deprecated JDK method.


common/src/main/java/com/tc/async/impl/StageMonitorImpl.java
```diff
@@ -1,160 +0,0 @@
...
- */
-package com.tc.async.impl;
-
-import com.tc.async.api.StageMonitor;
-import com.tc.text.StringFormatter;
-
-import java.util.ArrayList;
-import java.util.List;
-
-class StageMonitorImpl implements StageMonitor {
```
chrisdennis, line 26: Was this unused?

chrisdennis (approved):
Bunch of unrelated changes... but they all seem reasonable.

common/src/main/java/com/tc/async/impl/StageMonitorImpl.java
myronkscott: yep

myronkscott: removed the interface. and all implementations.  none of it is used

myronkscott:
> Bunch of unrelated changes... but they all seem reasonable.

I'll separate commits

common/src/main/java/com/tc/async/impl/StageMonitorImpl.java
devendra-ibm: is `org.terracotta.tripwire.StageMonitor` somewhat replacement for this .?

devendra-ibm (approved):

PR #1432, opened 2026-09-15
Replica and Relay need start and connected states

myronkscott:
Both Replica and Relay need start and connected states so each can be differentiated from having no data and data that must be validated or erased on connection.

Fix for connected clients on passives tracking included.

Election changes to handle Replica with peers.


common/src/main/java/com/tc/util/SimpleRingBuffer.java
```diff
@@ -28,21 +28,30 @@
...
   public SimpleRingBuffer(int size) {
     buffer = new Object[size];
     head = 0;
   }
-  
+
+  public void clear() {
+    Arrays.setAll(buffer, (value) -> null);
+    head = 0;
+  }
```
guptakshit: There is another method `fill(Object[] a, Object val)` which avoids creating lambdas

common/src/main/java/com/tc/util/SimpleRingBuffer.java
```diff
@@ -28,21 +28,30 @@
...
+
+  public boolean isEmpty() {
+    return head == 0 && Arrays.stream(buffer).allMatch(Objects::isNull);
+  }
+
```
guptakshit: For every isEmpty() check it'll iterate over all 5000 objects, would it better to maintain an explicitly `size` variable to avoid iterating all objects or built a custom circular buffer using java Deque?

tc-server/src/main/java/com/tc/objectserver/handler/DuplicationTransactionHandler.java
```diff
@@ -146,6 +144,9 @@ private void processMessage(Stage<ReplicationMessage> sendToNext, RelayMessage r
             waitFors.put(a.getActivityID(), new CompletableFuture<>());
           }
         }
+        if (currentSequence+1 != msg.getSequenceID()) {
+          LOGGER.warn("sequence skipped current: {} message: {}", currentSequence, msg.getSequenceID());
+        }
         currentSequence = Long.max(currentSequence, msg.getSequenceID());
         sendToNext.getSink().addToSink(msg);
```
guptakshit: if the replica sees gap in ids and loss of data should it throw `ZapDirtyDbServerNodeException` instead of log or it can never happen (in that case assertion error is thrown)?

tc-server/src/main/java/com/tc/l2/state/ElectionManagerImpl.java
```diff
@@ -300,7 +300,9 @@ private NodeID doElection(NodeID myNodeId, Set<String> servers, boolean isNew, W
       Assert.assertEquals(msg.getMessageID(), response.inResponseTo());
       if (response.getType() == L2StateMessage.RESULT_AGREED) {
         Assert.assertEquals(e, response.getEnrollment());
-        if (StateManager.convert(response.getState()) == ServerMode.PASSIVE) {
+        if (StateManager.convert(response.getState()) == ServerMode.RELAY_CONNECTED) {
+          passives.add(response.messageFrom());
+        } else if (StateManager.convert(response.getState()) == ServerMode.PASSIVE) {
```
guptakshit: Is this change to fix the case where a newly elected Active does not send messages to the Relay server? If the server is in Relay state (but not Relay-Connected), would that path be handled when a Replica connects, the Relay will request a sync from the Active?

tc-server/src/main/java/com/tc/objectserver/handler/RelayTransactionHandler.java
```diff
@@ -48,20 +50,39 @@ public class RelayTransactionHandler {
...
+    this.listener = new GroupEventsListener() {
+      @Override
+      public void nodeJoined(NodeID nodeID) {
+
+      }
+
+      @Override
+      public void nodeLeft(NodeID nodeID) {
+        if (nodeID.equals(endTarget)) {
+          clearForward();
```
guptakshit: now when the Replica leaves, the Relay will stop sending it the messages and it'll prevent Relay from sending wrong value when the Replica joins back?

tc-server/src/main/java/com/tc/objectserver/handler/RelayTransactionHandler.java
```diff
@@ -126,36 +155,59 @@ private void processMessage(ReplicationMessage rep) throws ServerException {
...
+    this.forward = batcher;
+    this.activeID = (ServerID)stateMgr.getActiveNodeID();
+    return valid;
   }
-  
-  private void sendToRelayTarget() {
+
+  private synchronized void clearHistory() {
+    this.activeID = null;
+    generationOffset = this.history.stream().mapToLong(ReplicationMessage::getSequenceID).max().orElse(0L);
```
guptakshit: If the history is empty it'll set `generationOffset` to 0 but the starting value is set to `  private long generationOffset = 1L;` wouldn't it cause mismatch of `currentSequence+1` and `msg.getSequenceId()` and relay sequence id will always 1 behind of replica sequence id?

common/src/main/java/com/tc/util/SimpleRingBuffer.java
guptakshit, line 43: There is fill(Object[] a, Object val) which avoids creating lambdas

common/src/main/java/com/tc/util/SimpleRingBuffer.java
guptakshit, line 48: For every isEmpty() check it'll iterate over all 5000 objects, would it better to maintain an explicit size variable to avoid iterating all objects?

tc-server/src/main/java/com/tc/l2/state/ElectionManagerImpl.java
guptakshit, line 305: Is this change to fix the case where a newly elected Active does not send messages to the Relay server?

But if the server is in Relay state (not Relay-Connected), would that path be handled when a fresh Replica connects and the Relay will request a sync from the Active?

tc-server/src/main/java/com/tc/objectserver/handler/DuplicationTransactionHandler.java
```diff
@@ -146,6 +144,9 @@ private void processMessage(Stage<ReplicationMessage> sendToNext, RelayMessage r
             waitFors.put(a.getActivityID(), new CompletableFuture<>());
           }
         }
+        if (currentSequence+1 != msg.getSequenceID()) {
+          LOGGER.warn("sequence skipped current: {} message: {}", currentSequence, msg.getSequenceID());
+        }
```
guptakshit, line 151: if the replica sees gap in IDs that means there's loss of data, should it throw ZapDirtyDbServerNodeException instead of log message or this gap can never happen (in that case assertion error is appropriate)?

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
```diff
@@ -665,17 +682,34 @@ private void handleElectionAbort(L2StateMessage clusterMsg) {
...
+    ServerMode other = StateManager.convert(clusterMsg.getState());
+    if (other == ServerMode.REPLICA || other == ServerMode.REPLICA_START) {
+      consoleLogger.info("Replica designated in cluster.  No Election will be performed");
+      sendVerificationOKResponse(clusterMsg);
+      setActiveNodeID(clusterMsg.messageFrom());
+      throw new TCShutdownServerException("There is a REPLICA member of this stripe.  Shutting down.");
+    } else {
+      verifyActiveDeclarationAndRespond(clusterMsg);
+    }
   }
```
guptakshit, line 682: From our discussions, weren't the other servers with replicas in the same stripe expected to remain running and transition to a diagnostic (or uninitialized) state?
Is there a reason we're shutting down other servers or design changed?

tc-server/src/main/java/com/tc/l2/state/ServerMode.java
```diff
@@ -114,6 +115,32 @@ public boolean canBeActive() {
...
+
+    @Override
+    public boolean canStartElection() {
+      return false;
+    }
+
+    @Override
+    public boolean containsData() {
+      return true;
+    }
```
guptakshit: There are benefits of having a different state which can be used to distinguish a RELAY which is connected to Peer vs which is not. But I didn't understand is the introduction of change for improvement or some of the functionality couldn't be achieved without having RELAY-CONNECTED because in `RelayPersistentState` is always returning 

```java
  @Override
  public ServerMode getInitialMode() {
    return ServerMode.RELAY;
  }
```

tc-server/src/main/java/com/tc/l2/state/ServerMode.java
```diff
@@ -114,6 +115,32 @@ public boolean canBeActive() {
...
       return false;
     }
+  },
+    RELAY_CONNECTED(PASSIVE_RELAY_CONNECTED) {
+    @Override
+    public boolean isStartup() {
+      return false;
+    }
+
+    @Override
```
guptakshit, line 125: RELAY-CONNECTED is distinct state to differentiate a RELAY that is connected to a peer from one that isn't. I'm wondering whether it was introduced mainly for clarity and improved state management, or if there is some functionality that couldn't be achieved without it?

tc-server/src/main/java/com/tc/server/TCServerImpl.java
```diff
@@ -509,6 +509,10 @@ public boolean replicaFailoverToActive() {
       dsoServer.getConfigSetupManager().ignoreReplicaSettings();
       leaveGroup();
       return true;
+    } if (getStateManager().getCurrentMode() == ServerMode.REPLICA_START) {
+      dsoServer.getConfigSetupManager().ignoreReplicaSettings();
+      leaveGroup();
+      return true;
     } else {
```
guptakshit, line 516: An unsynced replica (replica-start) can now be transitioned to Active. Was there a reason this transition wasn't allowed before?

tc-server/src/main/java/com/tc/objectserver/handler/RelayTransactionHandler.java
myronkscott: I don't think it is a problem because there is no previous history to mis-match but it should be consistent.  Will fix.

tc-server/src/main/java/com/tc/l2/state/ElectionManagerImpl.java
myronkscott: This occurs when a new Active takes over and the RELAY-CONNECTED is still in sync.  This allows the RELAY-CONNECTED and REPLICA continue on when this failover happens.  If the server is on RELAY, the RELAY will just wit until the REPLICA connects to request a new sync.

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
myronkscott: This path will not happen in production as the implementation currently stands.  Currently the group manager will prevent anything other than the RELAY from connecting to the REPLICA so this election will not occur.  It's dead code right now but I am still working out if we can use it so I left the implementation.  I have a separate item to put any additional servers in the stripe in diagnostic mode via dynamic-config.

tc-server/src/main/java/com/tc/server/TCServerImpl.java
myronkscott: If a server goes down in REPLICA mode and restarts from persistent data.  We want to be able to transition to active if requested.

common/src/main/java/com/tc/util/SimpleRingBuffer.java
myronkscott: This is rarely called and only for an assertion check.  I think this is best to keep the implementation simple without complicating a new variable and attempting to keep it in sync.

tc-server/src/main/java/com/tc/l2/state/ServerMode.java
myronkscott: It is to differentiate the connected state that holds data (history) vs. a fresh server that does not hold any data.

tc-server/src/main/java/com/tc/objectserver/handler/DuplicationTransactionHandler.java
myronkscott: That's a reasonable change but if this happens we have a real implementation error so we should look for this aggressively in test logs.

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
```diff
@@ -355,8 +356,11 @@ private void moveToPassiveReady(L2StateMessage src) {
     Enrollment verify = getVerificationEnrollment();
     setActiveNodeID(active);
     if (startState == ServerMode.RELAY) {
-      setActiveNodeID(active);
-      switchToState(ServerMode.RELAY, EnumSet.of(ServerMode.INITIAL, ServerMode.START, ServerMode.RELAY));
+      if (state == ServerMode.RELAY_CONNECTED) {
+        switchToState(ServerMode.RELAY_CONNECTED, EnumSet.of(ServerMode.INITIAL, ServerMode.START, ServerMode.RELAY, ServerMode.RELAY_CONNECTED));
```
devendra-ibm: `RELAY_CONNECTED ` state transition will always follow this pattern right .?
`INITIAL -> RELAY -> RELAY_CONNECTED`
though we have already check in outer loop, but it would be better if we tighten the valid states constraint here as well.

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
```diff
@@ -423,12 +427,25 @@ public void moveToDiagnosticMode() {
...
+    if (startState.containsData()) {
+      zapAndResyncLocalNode("relay contains data");
+    } else {
+      switchToState(ServerMode.RELAY, EnumSet.of(ServerMode.INITIAL, ServerMode.RELAY, ServerMode.RELAY_CONNECTED));
+    }
+  }
+
+  @Override
+  public void moveToRelayConnectedMode() {
+    if (startState.containsData()) {
```
devendra-ibm: I didn't understand this, if the intent is that server was running previously in `RELAY_CONNECTED` then it must have replicated some data to Replica, So it should zap itself, So that replica can restart itself as fresh instead of syncing in old data-dir.
but then we should persist `RELAY_CONNECTED` state as well, as of now i see we're reading in-memory via  `RelayPersistentState` class which is always `RELAY`

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
```diff
@@ -423,12 +427,25 @@ public void moveToDiagnosticMode() {
 
   @Override
   public void moveToRelayMode() {
-      switchToState(ServerMode.RELAY, EnumSet.of(ServerMode.INITIAL, ServerMode.RELAY));
+    if (startState.containsData()) {
```
devendra-ibm: Not sure about the significance of this check if we're reading in-memory state via `RelayPersistentState` class which is always `RELAY`, so this condition will always false .?

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
```diff
@@ -355,8 +356,11 @@ private void moveToPassiveReady(L2StateMessage src) {
     Enrollment verify = getVerificationEnrollment();
     setActiveNodeID(active);
     if (startState == ServerMode.RELAY) {
-      setActiveNodeID(active);
-      switchToState(ServerMode.RELAY, EnumSet.of(ServerMode.INITIAL, ServerMode.START, ServerMode.RELAY));
+      if (state == ServerMode.RELAY_CONNECTED) {
```
devendra-ibm: if it's in already RELAY_CONNECTED when new Active takes over, So the below
 `switchToState(ServerMode.RELAY_CONNECTED, EnumSet.of(ServerMode.INITIAL, ServerMode.START, ServerMode.RELAY, ServerMode.RELAY_CONNECTED));` code  is no-op right , So why we even need this .?

tc-server/src/main/java/com/tc/objectserver/handler/RelayTransactionHandler.java
devendra-ibm, line 77: Can this be possible now, if relay receives nodeLeft event for replica, then it will make forward null, so no further processing, but what if before it sets null, it has processed some messages(n+1, n+2) and send to replica, msg is received but not processed yet (i.e. might be lying somewhere in replica stage queue), and meanwhile connection heals and replica uses resume message  from current sequence n, while now n+1 and n+2 msgs gets processed. So since replica resumes from n, So there is possibility of duplicate msgs(n+1, n+2) arriving on replica. though it will restart the replica as now we ‘re throwing zap exception , but should this be more tightened at replica layer as well. meaning before creating resume we process left over messages if any, and then ask to resume. can this happen in very unstable environment , though it’s very tiny window.?

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
myronkscott: correct

myronkscott: Start and relay connected are both possible here.

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
myronkscott: This is correct.  It is a useless check.

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
myronkscott: This is correct.  It is a useless check.

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
myronkscott: It is essentially noop but for completeness.  I guess I can just remove the switches in relay mode.

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
devendra-ibm: i didn't understand how Start is possible here.?

tc-server/src/main/java/com/tc/objectserver/handler/RelayTransactionHandler.java
myronkscott: That is an interesting thought experiment.  I think we are protected by the currentSequence validation in the DuplicateTransactionHandler.

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
myronkscott: Code is removed but it was possible once election is started.

tc-server/src/main/java/com/tc/l2/state/StateManagerImpl.java
devendra-ibm: okay.

devendra-ibm (approved):

guptakshit (approved):

